PR #57, opened 2024-04-22
Default appPathPrefix to empty string

davidjamesstone:
Replace the default `appPathPrefix` (/forms-runner) with empty string.

Keep the capabilty to have a path prefix in case it needs to be reintroduced.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -84,7 +84,9 @@ const serverOptions = (): ServerOptions => {
 }
 
 const registrationOptions = {
-  routes: { prefix: config.appPathPrefix }
+  routes: {
+    prefix: config.appPathPrefix === '' ? undefined : config.appPathPrefix
+  }
 }
 
 async function createServer(routeConfig: RouteConfig) {
--- a/src/server/plugins/engine/plugin.ts
+++ b/src/server/plugins/engine/plugin.ts
@@ -79,11 +79,11 @@ export const plugin = {
     server.app.forms = {}
     const forms = server.app.forms
 
-    configs.forEach((config) => {
+    configs.forEach((form) => {
       const basePath =
-        formBasePath === '' ? config.id : `${formBasePath}/${config.id}`
+        formBasePath === '' ? form.id : `${formBasePath}/${form.id}`
 
-      forms[config.id] = new FormModel(config.configuration, {
+      forms[form.id] = new FormModel(form.configuration, {
         ...modelOptions,
         basePath
       })
--- a/src/server/plugins/engine/plugin.ts
+++ b/src/server/plugins/engine/plugin.ts
@@ -80,9 +80,12 @@ export const plugin = {
     const forms = server.app.forms
 
     configs.forEach((config) => {
+      const basePath =
+        formBasePath === '' ? config.id : `${formBasePath}/${config.id}`
+
       forms[config.id] = new FormModel(config.configuration, {
         ...modelOptions,
-        basePath: `${formBasePath}/${config.id}`
+        basePath
       })
     })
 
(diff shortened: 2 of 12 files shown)


src/server/plugins/engine/plugin.ts
```diff
@@ -80,9 +80,12 @@ export const plugin = {
     const forms = server.app.forms
 
     configs.forEach((config) => {
+      const basePath =
+        formBasePath === '' ? config.id : `${formBasePath}/${config.id}`
```
colinrotherham, line 84: Could we use POSIX `join()` here to always do forward slashes?

```patch
-import { join } from 'node:path'
+import { join } from 'node:path/posix'
```

I think it skips the `/` when one of the path segments is empty

```suggestion
      const basePath = join(formBasePath, config.id)
```

davidjamesstone: Agree we probably shouldn't be string concating URLs, but I'm not sure we should be using `node:path` for URLs, even the `posix` variant.

There are other (file path) uses of the imported `join` in that file too, which I wouldn't want to break for Windows users.

colinrotherham: Can add a whole new import for `posix.join()`?

Seemed like a nice way to handle **appPathPrefix** as `''` but it also works with  `'/'`

```patch
import { join } from 'node:path'
+import posix from 'node:path/posix'
```

```suggestion
      const basePath = posix.join(formBasePath, config.id)
```

davidjamesstone: Your note about `''` and `'/'` made me think again about this. Initially I wanted to default `appPathPrefix` to undefined but there were assumptions in the code that it would be a string. I have removed those assumptions now and introduced an `appPathPrefixer` function that is now used in `js` and `njk` files. I've gone with `posix.join` for now (but don't think it should be used for building urls generally). I have also applied a joi schema to the `appPathPrefix` configuration key to assert it is a valid prefix (it now must be a valid relative uri, empty string and `'/'` are disallowed)

src/server/index.ts
```diff
@@ -84,7 +84,9 @@ const serverOptions = (): ServerOptions => {
 }
 
 const registrationOptions = {
-  routes: { prefix: config.appPathPrefix }
+  routes: {
+    prefix: config.appPathPrefix === '' ? undefined : config.appPathPrefix
```
colinrotherham, line 88: Can probably let the `''` fall back to `undefined`?

Then we only have to remove the ` || undefined` in your improved PR

```suggestion
    prefix: config.appPathPrefix || undefined
```

src/server/plugins/engine/plugin.ts
colinrotherham, line 84: Another suggestion, fewer edits to make in your next PR

```suggestion
      const basePath = formBasePath ? `${formBasePath}/${config.id}` : config.id
```

src/server/plugins/engine/plugin.ts
```diff
@@ -80,9 +80,12 @@ export const plugin = {
     const forms = server.app.forms
 
     configs.forEach((config) => {
```
colinrotherham, line 82: This caught me out a bit, having a clash with `config` (import) versus `config` scoped

Should we rename it?

```suggestion
    configs.forEach((form) => {
```

colinrotherham (approved):
Thanks @davidjamesstone

Approved! Added some optional things, see what you think

src/server/plugins/engine/plugin.ts
davidjamesstone: Yeah this caught me out too. I had previously renamed the local forEach variable in the earlier PR.

davidjamesstone: Resolved https://github.com/DEFRA/forms-runner/pull/57/commits/96c9c1fff40755b20a0201e94032f52841ad43d6

PR #65, opened 2024-04-26
Fix/348784 undefined condition evaluation

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/348784

Fix incorrect routing when (negative) condition evaluation involves optional Checkbox and/or Radio component.

### DIFF
--- a/src/server/plugins/engine/pageControllers/PageControllerBase.ts
+++ b/src/server/plugins/engine/pageControllers/PageControllerBase.ts
@@ -393,15 +393,18 @@ export class PageControllerBase {
       for (const component of nextPage.components.items) {
         let componentState = currentState[component.name]
 
-        // For evaluation context purposes, optional `CheckboxesField`s
-        // with an undefined value (i.e. nothing selected) should default to [].
-        // This way conditions are not evaluated against `undefined` which throws errors (see LN234).
-        // Currently these errors are caught and the evaluation returns default `false`.
-        // (See src/server/plugins/engine/models/FormModel.ts LN227)
-        // For negative conditions this is a problem because E.g.
-        // The condition: 'selectedchecks' does not contain 'someval'
-        // should return true IF 'selectedchecks' is undefined, not throw and return false.
-        // Similarly for optional `RadiosField`, the evaluation context should default to null.
+        /**
+         * For evaluation context purposes, optional {@link CheckboxesField}
+         * with an undefined value (i.e. nothing selected) should default to [].
+         * This way conditions are not evaluated against `undefined` which throws errors.
+         * Currently these errors are caught and the evaluation returns default `false`.
+         * @see {@link PageControllerBase.getNextPage} for `undefined` return value
+         * @see {@link FormModel.makeCondition} for try/catch block with default `false`
+         * For negative conditions this is a problem because E.g.
+         * The condition: 'selectedchecks' does not contain 'someval'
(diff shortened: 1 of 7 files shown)


test/cases/server/plugins/engine/conditions/checkboxes.test.js
```diff
@@ -0,0 +1,78 @@
...
+import FormData from 'form-data'
+
+const testDir = dirname(fileURLToPath(import.meta.url))
+const okStatusCode = 200
+const redirectStatusCode = 302
+const htmlContentType = 'text/html'
+const key = 'wqJmSf'
+
+describe('Checkboxes based conditions', () => {
+  let server
```
colinrotherham, line 13: Not a blocker, but these new tests are great and deserve to be ESLint warning-free 🙌

```suggestion
  /** @type {import('@hapi/hapi').Server} */
  let server
```

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -389,7 +391,32 @@ export class PageControllerBase {
...
+
+        // For evaluation context purposes, optional `CheckboxesField`s
+        // with an undefined value (i.e. nothing selected) should default to [].
+        // This way conditions are not evaluated against `undefined` which throws errors (see LN234).
+        // Currently these errors are caught and the evaluation returns default `false`.
+        // (See src/server/plugins/engine/models/FormModel.ts LN227)
+        // For negative conditions this is a problem because E.g.
+        // The condition: 'selectedchecks' does not contain 'someval'
+        // should return true IF 'selectedchecks' is undefined, not throw and return false.
+        // Similarly for optional `RadiosField`, the evaluation context should default to null.
```
colinrotherham, line 404: Thanks @davidjamesstone, this write up is really clear

Mind if we use JSDoc `@link` to replace line numbers that can drift?

```suggestion
        /**
         * For evaluation context purposes, optional {@link CheckboxesField}
         * with an undefined value (i.e. nothing selected) should default to [].
         * This way conditions are not evaluated against `undefined` which throws errors.
         * Currently these errors are caught and the evaluation returns default `false`.
         * @see {@link PageControllerBase.getNextPage} for `undefined` return value
         * @see {@link FormModel.makeCondition} for try/catch block with default `false`
         * For negative conditions this is a problem because E.g.
         * The condition: 'selectedchecks' does not contain 'someval'
         * should return true IF 'selectedchecks' is undefined, not throw and return false.
         * Similarly for optional {@link RadiosField}, the evaluation context should default to null.
         */
```

colinrotherham (approved):

test/cases/server/plugins/engine/conditions/checkboxes.test.js
davidjamesstone: [Resolved](https://github.com/DEFRA/forms-runner/pull/65/commits/355a74194b0e753f0a5155b16dbfff5dc3ab41a6)

src/server/plugins/engine/pageControllers/PageControllerBase.ts
davidjamesstone: [Resolved](https://github.com/DEFRA/forms-runner/pull/65/commits/7af8bdd5f691ef2e08294784b2a6a0c5380a7c12)

alexluckett (approved):

PR #94, opened 2024-05-08
Feature/369420 run forms from db

davidjamesstone:
Parts of the PR look a little messy because I am trying to keep the capability to run the server using a `.json` file for now.
There's quite a bit of other code that can be removed now but I have kept it in for the time being as there are some useful tests around it.

I plan to come back and remove the following:

- Preview mode
- Ability to run a json file
- The json files in `src/server/forms`
- The `/publishing` [endpoints](https://github.com/DEFRA/forms-runner/blob/main/src/server/plugins/engine/plugin.ts#L103)
- The forms-runner [authentication](https://github.com/DEFRA/forms-runner/blob/main/src/server/plugins/auth.ts) capabilty

@alexluckett Am I right in thinking all this can go?

### DIFF
--- a/src/server/plugins/engine/configureEnginePlugin.ts
+++ b/src/server/plugins/engine/configureEnginePlugin.ts
@@ -16,7 +16,7 @@ type ConfigureEnginePlugin = (
 ) => {
   plugin: any
   options: {
-    model: any
+    model?: FormModel
     previewMode: boolean
   }
 }
--- a/src/server/plugins/engine/configureEnginePlugin.ts
+++ b/src/server/plugins/engine/configureEnginePlugin.ts
@@ -32,16 +32,17 @@ export const configureEnginePlugin: ConfigureEnginePlugin = (
   // let configs: FormConfiguration[]
   let model
 
-  // const modelOptions = {
-  //   relativeTo: join(config.appDir, 'plugins/engine/views'),
-  //   previewMode: options?.previewMode ?? config.previewMode
-  // }
+  const modelOptions = {
+    relativeTo: join(config.appDir, 'plugins/engine/views'),
+    previewMode: options?.previewMode ?? config.previewMode
+  }
 
   if (formFileName && formFilePath) {
     const formConfigPath = join(formFilePath, formFileName)
     const definition = JSON.parse(fs.readFileSync(formConfigPath, 'utf8'))
     model = new FormModel(definition, {
-      basePath: idFromFilename(formFileName)
+      basePath: idFromFilename(formFileName),
+      ...modelOptions
     })
     //   configs = [
     //     {
--- a/src/server/plugins/engine/configureEnginePlugin.ts
+++ b/src/server/plugins/engine/configureEnginePlugin.ts
@@ -30,7 +30,7 @@ export const configureEnginePlugin: ConfigureEnginePlugin = (
   options?: EngineOptions
 ) => {
   // let configs: FormConfiguration[]
-  let model
+  let model: FormModel | undefined
 
   const modelOptions = {
     relativeTo: join(config.appDir, 'plugins/engine/views'),
--- a/src/server/plugins/engine/configureEnginePlugin.ts
+++ b/src/server/plugins/engine/configureEnginePlugin.ts
@@ -3,11 +3,12 @@ import { join } from 'node:path'
 
 import config from '~/src/server/config.js'
 import { idFromFilename } from '~/src/server/plugins/engine/helpers.js'
+import { FormModel } from '~/src/server/plugins/engine/models/FormModel.js'
 import { plugin } from '~/src/server/plugins/engine/plugin.js'
-import {
-  loadPreConfiguredForms,
-  type FormConfiguration
-} from '~/src/server/plugins/engine/services/configurationService.js'
+// import {
+//   loadPreConfiguredForms,
+//   type FormConfiguration
+// } from '~/src/server/plugins/engine/services/configurationService.js'
 
 type ConfigureEnginePlugin = (
   formFileName?: string,
@@ -15,14 +16,7 @@ type ConfigureEnginePlugin = (
 ) => {
   plugin: any
   options: {
-    modelOptions: {
-      relativeTo: string
-      previewMode: any
-    }
-    configs: {
-      configuration: any
-      id: string
-    }[]
+    model: any
     previewMode: boolean
   }
 }
@@ -35,28 +29,32 @@ export const configureEnginePlugin: ConfigureEnginePlugin = (
   formFilePath,
   options?: EngineOptions
 ) => {
-  let configs: FormConfiguration[]
+  // let configs: FormConfiguration[]
+  let model
+
+  // const modelOptions = {
+  //   relativeTo: join(config.appDir, 'plugins/engine/views'),
+  //   previewMode: options?.previewMode ?? config.previewMode
+  // }
 
   if (formFileName && formFilePath) {
     const formConfigPath = join(formFilePath, formFileName)
-
-    configs = [
-      {
-        configuration: JSON.parse(fs.readFileSync(formConfigPath, 'utf8')),
-        id: idFromFilename(formFileName)
-      }
-    ]
-  } else {
-    configs = loadPreConfiguredForms()
-  }
-
-  const modelOptions = {
-    relativeTo: join(config.appDir, 'plugins/engine/views'),
-    previewMode: options?.previewMode ?? config.previewMode
+    const definition = JSON.parse(fs.readFileSync(formConfigPath, 'utf8'))
+    model = new FormModel(definition, {
+      basePath: idFromFilename(formFileName)
+    })
+    //   configs = [
+    //     {
+    //       configuration: JSON.parse(fs.readFileSync(formConfigPath, 'utf8')),
+    //       id: idFromFilename(formFileName)
+    //     }
+    //   ]
+    // } else {
+    //   configs = loadPreConfiguredForms()
   }
 
   return {
     plugin,
-    options: { modelOptions, configs, previewMode: config.previewMode }
+    options: { model, previewMode: config.previewMode }
   }
 }
--- a/src/server/plugins/engine/plugin.ts
+++ b/src/server/plugins/engine/plugin.ts
@@ -73,6 +73,10 @@ function getStartPageRedirect(
 
 interface PluginOptions {
   relativeTo?: string
+  modelOptions: {
+    relativeTo: string
+    previewMode: any
+  }
   model: any
   previewMode: boolean
 }
@@ -82,7 +86,7 @@ export const plugin = {
   dependencies: '@hapi/vision',
   multiple: true,
   register: (server: Server, options: PluginOptions) => {
-    const { model, previewMode } = options
+    const { model, previewMode, modelOptions } = options
     const enabledString = config.previewMode ? `[ENABLED]` : `[DISABLED]`
     const disabledRouteDetailString =
       'A request was made however previewing is disabled. See environment variable details in runner/README.md if this error is not expected.'
@@ -260,7 +264,10 @@ export const plugin = {
             server.logger.info(
               `Building model for form definition ${id} (${slug})`
             )
-            const model = new FormModel(definition, { basePath: slug })
+            const model = new FormModel(definition, {
+              basePath: slug,
+              ...modelOptions
+            })
 
             // Create new cache item
             item = { model, updatedAt: state.updatedAt }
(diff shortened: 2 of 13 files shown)


src/server/plugins/engine/plugin.ts
```diff
@@ -77,21 +86,25 @@ export const plugin = {
   dependencies: '@hapi/vision',
   multiple: true,
   register: (server: Server, options: PluginOptions) => {
-    const { modelOptions, configs, previewMode } = options
-    server.app.forms = {}
-    const forms = server.app.forms
```
colinrotherham, line 82: Ah this `const forms` is used a few times further down still

The routes `'/publish'`, `'/published/{id}'` and `'/published'` try to use `forms[id]`

src/server/plugins/engine/plugin.ts
```diff
@@ -192,63 +205,118 @@ export const plugin = {
...
-      handler: (request: Request, h: ResponseToolkit) => {
-        const { id } = request.params
-        const model = forms[id]
-        if (model) {
-          return getStartPageRedirect(request, h, id, model)
+        ],
+        validate: {
+          params: Joi.object().keys({
+            slug: Joi.string().required()
+          })
```
colinrotherham, line 305: One for `formBySlugSchema` now we export it?

test/cases/server/plugins/auth.test.ts
```diff
@@ -37,7 +42,7 @@ describe('Server Auth', () => {
       )
     })
 
-    test('sign in page returns to teh previous url', async () => {
+    test('sign in page returns to the previous url', async () => {
```
colinrotherham, line 45: 😎

src/server/plugins/engine/plugin.ts
davidjamesstone: I'll use `slugSchema` from `@defra/forms-model` for now. `formBySlugSchema` lives in `forms-manager` at the moment (but should probably be promoted to `@defra/forms-model` to allow for reuse).

davidjamesstone: Resolved with 002344074e0211dd1ff9b97fc07627d77adbb45c

src/server/plugins/engine/plugin.ts
davidjamesstone: Resolved with 6760ebe607166cd42998a2fc12ce545a0ad9b634

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,39 @@
...
+import config from '~/src/server/config.js'
+import { getJson } from '~/src/server/plugins/engine/services/httpService.js'
+
+const { managerUrl } = config
+
+/**
+ * Retrieves a form definition from the form manager for a given slug
+ * @param {string} slug - the slug of the form
+ */
+export async function getFormMetadata(slug) {
```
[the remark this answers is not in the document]
davidjamesstone: Initially I pulled in the `fetch.js` from `forms-designer`, with all the new `URL` goodness you added.
However, I reverted to using the pre-existing `import { getJson } from '~/src/server/plugins/engine/services/httpService.js'`. This is because there's a subtle differences in `Wreck.request(method, url)` vs `Wreck.get(url)`. These shorthand methods have [some useful error handling](https://github.com/hapijs/wreck/blob/master/lib/index.js#L547) built-in.

I then spent too long trying to get the types to play nicely with the shorthand methods until I eventually gave in and used what I already had at my disposal. I would be great to try and get it working together if we get a free hour (including maybe moving to native `fetch`).

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,39 @@
+import { formMetadataSchema } from '@defra/forms-model'
+
+import config from '~/src/server/config.js'
+import { getJson } from '~/src/server/plugins/engine/services/httpService.js'
```
colinrotherham, line 4: I get what you mean about the types in this one. Does anything else use it?

Probably best sticking with one and I'll look at the types in another PR

```suggestion
import { getJson } from '~/src/server/services/httpService.js'
```

src/server/plugins/engine/configureEnginePlugin.ts
```diff
@@ -35,28 +29,33 @@ export const configureEnginePlugin: ConfigureEnginePlugin = (
...
-      {
-        configuration: JSON.parse(fs.readFileSync(formConfigPath, 'utf8')),
-        id: idFromFilename(formFileName)
-      }
-    ]
-  } else {
-    configs = loadPreConfiguredForms()
-  }
+  // let configs: FormConfiguration[]
+  let model
```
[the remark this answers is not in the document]
davidjamesstone: Resolved with 8f0154b13d62699ab89cfa6b87778826839f9fd8

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,39 @@
...
+  const metadata = await getJson(`${managerUrl}/forms/slug/${slug}`)
+
+  // Run it through the schema to coerce dates
+  const { value, error } = formMetadataSchema.validate(metadata)
+
+  if (error) {
+    throw error
+  }
+
+  return value
```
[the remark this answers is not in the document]
davidjamesstone: Resolved with 8f0154b13d62699ab89cfa6b87778826839f9fd8

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Retrieves a form definition from the form manager for a given id
+ * @param {string} id - the id of the form
+ * @param {'draft'|'live'} state - the state of the form
+ */
+export async function getFormDefinition(id, state) {
+  const suffix = state === 'draft' ? '/draft' : ''
+  const defintion = await getJson(
+    `${managerUrl}/forms/${id}/definition${suffix}`
+  )
```
colinrotherham, line 34: I've fixed the types for the HTTP service in https://github.com/DEFRA/forms-runner/pull/95

It means you can now correctly type the payload, but it's a nested property unlike the plugin version of the HTTP service. Any good?

```suggestion
  const getJsonByType = /** @type {typeof getJson<FormDefinition>} */ (getJson)

  const suffix = state === 'draft' ? '/draft' : ''
  const { payload: definition } = await getJsonByType(
    `${managerUrl}/forms/${id}/definition${suffix}`
  )
```

src/server/plugins/engine/services/formsService.js
davidjamesstone: I'm going to stick with `~/src/server/plugins/engine/services/httpService.js` for now. I'm not sure why there are two `fetch` utils but there's subtly different behaviour between the two, namely this part:

```
    if (res.statusCode !== 200) {
      const err = payload || new Error('Unknown error')
      throw err
    }
```

Let's look at bringing over (or better still, making a shared library) the `URL` ones but typed against wrecks' shortcut methods `get`, `post` etc. This can be done in the next PR for the [clean up ticket](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/376156).

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,39 @@
...
+    throw error
+  }
+
+  return value
+}
+
+/**
+ * Retrieves a form definition from the form manager for a given id
+ * @param {string} id - the id of the form
+ * @param {FormState} state - the state of the form
```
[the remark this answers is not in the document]
davidjamesstone: Resolved with 8f0154b13d62699ab89cfa6b87778826839f9fd8

src/server/plugins/engine/services/formsService.js
colinrotherham: If you're sure?

I've fixed the types and resolved that subtle difference in:

* https://github.com/DEFRA/forms-runner/pull/95

src/server/plugins/engine/plugin.ts
```diff
@@ -100,155 +117,210 @@ export const plugin = {
...
+
       await cacheService.mergeState(request, newValues)
+
       return h.continue
     }
 
+    const loadFormPreHandler = server.app.model
+      ? async (request: Request, h: ResponseToolkit) =>
+          (request.app.model = server.app.model)
+      : async (request: Request, h: ResponseToolkit) => {
```
alexluckett, line 237: just a style preference, I find this syntax a bit difficult to read. 

Could this just be one function with a quick return at the top?

```javascript
const loadFormPreHandler = async (request: Request, h: ResponseToolkit) => {
    if(server.app.model) { 
       // do this bit
    }
   
    // else
}
```

src/server/plugins/engine/plugin.ts
```diff
@@ -100,155 +117,210 @@ export const plugin = {
...
       )
+
       await cacheService.mergeState(request, newValues)
+
       return h.continue
     }
 
+    const loadFormPreHandler = server.app.model
+      ? async (request: Request, h: ResponseToolkit) =>
+          (request.app.model = server.app.model)
```
alexluckett, line 236: not sure if I'm misreading, but this is only caching a single form at a time? What about the other forms?

src/server/plugins/engine/plugin.ts
davidjamesstone: It's only temporary and not what I'd usually do. At the time I thought we we're removing the capability to run a json file, but I wanted to retain the functionality while there are a bunch of tests that rely on it. If you don't like the ternary I can refactor now but it will all be getting re-done in the [clean-up ticket](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/376156).

src/server/plugins/engine/plugin.ts
davidjamesstone: `server.app.model` is only populated in the case where we're exclusively in "run from json file" mode, otherwise we run through the "real" `loadFormPrehandler` method. Inside there we cache models in the `itemsCache` using the key:

```
const key = `${id}_${formState}`
```
E.g. `663cc9a02095c9c7457806ff_draft`, `663cc9a02095c9c7457806ff_live`

[See here](https://github.com/DEFRA/forms-runner/blob/feature/369420-run-forms-from-db/src/server/plugins/engine/plugin.ts#L258)

src/server/plugins/engine/services/formsService.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Retrieves a form definition from the form manager for a given id
+ * @param {string} id - the id of the form
+ * @param {'draft'|'live'} state - the state of the form
+ */
+export async function getFormDefinition(id, state) {
+  const suffix = state === 'draft' ? '/draft' : ''
+  const defintion = await getJson(
```
colinrotherham, line 32: This one of yours @alexluckett?

```suggestion
  const definition = await getJson(
```

src/server/plugins/engine/services/formsService.js
colinrotherham: Here's a [branch](https://github.com/DEFRA/forms-runner/compare/feature/369420-run-forms-from-db...feature/369420-run-forms-from-db-rebased) with it all working in commit https://github.com/DEFRA/forms-runner/commit/851abff36eecb0f5f005107e886d9e064658daf8

Fixes all the ESLint errors in **formsService.js**

src/server/plugins/engine/plugin.ts
davidjamesstone: Resolved with 8d16917cb0bb396c2f281f89969c141c997cb1a7

src/server/plugins/engine/services/formsService.js
davidjamesstone: Resolved with ec8b633212ba5789795de182e833125535b35540

src/server/plugins/engine/plugin.ts
alexluckett: Aha, I understand. I was gettiing `request.app.model` and `server.app.model` mixed up.

Thanks for clarifying.

alexluckett: Aha, I understand. I was gettiing request.app.model and server.app.model mixed up.

Thanks for clarifying.

alexluckett (approved):

src/server/plugins/engine/services/formsService.js
davidjamesstone: Thanks Colin.

I will take this branch in once I've got this PR merged.
I was a little nervous doing it because of the other services currently using `src/server/services/httpService.ts`. i'm sure it's all fine but would like to do some testin/auditing first.

PR #105, opened 2024-05-14
Bugfix: view form bad request

alexluckett:
Since HTTPS wasn't used, forms were coming back empty and validation failed. This forces HTTPS for the default config entry, and bumps forms-model so validation passes with the new Azure IDs.


colinrotherham (approved):
Nice. Might as well do the same in **forms-designer**

PR #135, opened 2024-05-31
Fix application status so a confirmation dialog is displayed

alexluckett:
This was just a quick experiment to check the viability of using the existing output to notify/webhook code. We've since decided to go with a fresher approach, so I'm leaving this here as reference until that work is done and will delete this draft once done.


alexluckett:
Closing and deleting now Dave's work has completed.

PR #159, opened 2024-06-11
Set redis prefix upon connection

alexluckett:


src/server/plugins/session.ts
```diff
@@ -5,12 +5,16 @@ import config from '~/src/server/config.js'
...
       expiresIn: config.sessionTimeout
     },
+    storeBlank: false,
+    errorOnCacheNotReady: true,
     cookieOptions: {
       password: config.sessionCookiePassword,
-      isSecure: !!config.isDev,
+      isSecure: config.isProd,
       isHttpOnly: true,
       isSameSite: 'Lax'
```
colinrotherham, line 19: Just flagging that we don't have these two in **forms-designer**

src/server/plugins/session.ts
```diff
@@ -5,12 +5,16 @@ import config from '~/src/server/config.js'
 export default {
   plugin: yar,
   options: {
+    maxCookieSize: 0, // Always use server-side storage
     cache: {
+      cache: 'session',
       expiresIn: config.sessionTimeout
```
colinrotherham, line 11: Are we supposed to set `expiredIn` for **forms-designer** too?

colinrotherham (approved):

src/server/plugins/session.ts
alexluckett: We should do, I'll get those reviewed.

PR #167, opened 2024-06-13
Feature/393341 error page content

davidjamesstone:
Update error page content


src/server/views/500.html
```diff
@@ -6,7 +6,12 @@
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <h1 class="govuk-heading-l">Sorry, there is a problem with the service</h1>
-          <p class="govuk-body">Contact your closest consulate.</p>
+          <p class="govuk-body">You can:</p>
+          <ul class="govuk-list govuk-list--bullet">
+            <li>try your browser back button as your data may be saved</li>
```
alexluckett, line 11: should the "t" in "try" be capitalised, like how the "C" in "Contact" is?

colinrotherham: Lowercase is good because it's following the "You can" lead-in

>[**Bulleted lists**](https://design-system.service.gov.uk/styles/lists/#bulleted-lists)
>Introduce bulleted lists with a lead-in line ending in a colon. Start each item with a lowercase letter, and do not use a full stop at the end.

src/server/views/500.html
```diff
@@ -6,7 +6,12 @@
...
+            <li><a class="govuk-link" href="https://www.gov.uk/guidance/contact-defra">Contact the Defra Helpline</a> if you
```
colinrotherham, line 12:
```suggestion
            <li><a class="govuk-link" href="https://www.gov.uk/guidance/contact-defra">contact the Defra Helpline</a> if you
```

colinrotherham (approved):

PR #172, opened 2024-06-13
Update service name and the default serviceUrl

davidjamesstone:
- Remove serviceStartPage and serviceUrl from config
- Rename serviceStartPage to serviceUrl in code
- Default serviceName config to `Submit a form to Defra`


src/server/plugins/views.ts
```diff
@@ -72,7 +72,7 @@ export default {
       location: request?.app.location,
       sessionTimeout: config.sessionTimeout,
       skipTimeoutWarning: false,
-      serviceStartPage: config.serviceStartPage || '#',
+      serviceStartPage: config.serviceStartPage,
```
alexluckett, line 75: Can we remove the `config.serviceStartPage` entirely? I can see it gets set as `const serviceStartPage = `/${this.model.basePath}` in the PageControllerBase.ts which is good (so the links work in the forms), but we don't need that to be customisable by config.

davidjamesstone: Sure, I'll go ahead and remove it from confg too. I wasn't sure if it'd come in useful if/when we support running in "plugin" mode. Probably not.

davidjamesstone: Resolved with d694109cf3aa892c7e38bea3098feafa3d2d2734

(I renamed serviceStartPage to serviceUrl as that better reflects what it is and matches GDS)

PR #173, opened 2024-06-14
Add back link to summary page

davidjamesstone:


src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -104,6 +104,15 @@ export class SummaryPageController extends PageController {
       if (declarationError.length) {
         viewModel.declarationError = declarationError[0]
       }
+
+      const progress = state.progress ?? []
+
+      this.updateProgress(progress, request.url)
+
+      await cacheService.mergeState(request, { progress })
```
alexluckett, line 112: feels like this should sit within the `updateProgress` function, unless you can think of a use-case where updating the progress shouldn't also update the state? I feel like they're part of the same operation.

If that sounds sensible, we can remove the `request` parameter from `this.updateProgress` and make it something like:

```
  protected updateProgress(progress: string[], url: URL) {
    const progress = state.progress ?? []

    const lastVisited = progress.at(-1)
    const currentPath = `${this.model.basePath}${this.path}${url.search}`

    if (!lastVisited?.startsWith(currentPath)) {
      if (progress.at(-2) === currentPath) {
        progress.pop()
      } else {
        progress.push(currentPath)

        // Ensure progress history doesn't grow too large
        // by curtailing the array to max length of 100
        const MAX_PROGRESS_ENTRIES = 100

        if (progress.length > MAX_PROGRESS_ENTRIES) {
          progress.shift()
        }
      }
    }
    
    await cacheService.mergeState(request, { progress })
  }
```

alexluckett: maybe `pushProgress` is a more descriptive name than `updateProgress`?

davidjamesstone: Funnily enough, that's kind of how it was right until the final few minutes, the `updateProgress` (at the time called `setBackLink`) had both these lines in:

```
await cacheService.mergeState(request, { progress })

viewModel.backLink = progress.at(-2) ?? this.backLinkFallback
```

At which point I decided it was doing too much: updating and progress stack and setting the back link so I pulled it apart.

Thinking about it again, I can't see a reason we'd not want to persist. The method name `updateProgress` also suggests it should do!

We will need to pass in both the `cacheService` and `request` though.

davidjamesstone: Resolved with 4969fe101847bc9dc4c753f85b9d6ec98e0e0d2b
@alexluckett ?

alexluckett (approved):

src/server/plugins/engine/pageControllers/SummaryPageController.ts
alexluckett: Awesome, thanks 🙌

davidjamesstone: Thanks @alexluckett

PR #181, opened 2024-06-17
Feature/392297 help page

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/392297

Adds the help route, with capability to have both generic and form specific view modes.

Includes a (commented out) example approach on how we could load form specific help content using a nunjucks `switch` case statement.

If this statement becomes too big (as the number of forms increases), we could replace with a partial import e.g.

```
{% include "partials/help/{{form.id}}.html" ignore missing %}
```

Or we could start with this approach if we prefer?


src/server/plugins/router.ts
```diff
@@ -11,6 +12,24 @@ export default {
     register: (server) => {
       server.route(routes)
       server.route([
+        {
+          method: 'get',
+          path: '/help/{slug?}',
```
alexluckett, line 17: we have other pages underneath `/help/`. To avoid clashes with cookies, accessibility statement, etc., could we name this explicitly?

e.g. something like `/help/get-support/{slug?}`

src/server/views/help/index.html
```diff
@@ -0,0 +1,48 @@
...
+{% block content %}
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      {% if form %}
+        <h1 class="govuk-heading-l">Get help with this form</h1>
+      {% endif %}
+
+      {# Example pattern. Uncomment when we have some real form help content
+      {% switch form.id %}
```
davidjamesstone, line 16: We could replace this with something like:

`{% include "partials/help/{{form.id}}.html" ignore missing %}` 

And in the directory we would include the form-specific help partials e.g. `partials/help/66684c0e128c09b3a8444ce8.html`

[ignore missing](https://mozilla.github.io/nunjucks/templating#include) would protect us from missing content.

src/server/plugins/router.ts
davidjamesstone: I'd seen this as the `index` or root page of the help section. I'll change it to `/help/get-support/{slug}` :+1:

davidjamesstone: Resolved with d74a8b6f5e08c158f0277a96f3a712ca562d5830

src/server/views/help/index.html
davidjamesstone: Resolved with d74a8b6f5e08c158f0277a96f3a712ca562d5830

alexluckett (approved):

PR #183, opened 2024-06-18
Redirect to the current summary page path, rather than a hardcoded '/…

alexluckett:
- Redirect to the current summary page path, rather than a hardcoded `/summary` at the end of the form
- Set the start page of the form as the base path, because `/summary` isn't guaranted to exist and will show a 404 page in many cases as-is. Note I see this as a short term fix, we probably want a nicer UI solution for this (either fixing the summary paths or something like task list to help them get back on track?). This page shouldn't appear unless there's a bug anyway so the primary user journey won't be impacted.

### DIFF
--- a/src/server/plugins/engine/models/SummaryViewModel.ts
+++ b/src/server/plugins/engine/models/SummaryViewModel.ts
@@ -240,7 +240,7 @@ function Item(
   page,
   model: FormModel,
   params: { num?: number; returnUrl: string } = {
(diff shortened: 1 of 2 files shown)


nayij:
Can this be closed pls?

colinrotherham:
Thanks @nayij

Let's close this as summary pages only have `/summary` path since:

* https://github.com/DEFRA/forms-designer/pull/430

PR #184, opened 2024-06-18
Feature/382002 remove html5 validation

davidjamesstone:
- [Add novalidate to all forms](https://github.com/DEFRA/forms-runner/pull/184/commits/785ca0368a71a720c2f3b11de33afae97200f663)
- [Remove schema.max html attribute](https://github.com/DEFRA/forms-runner/pull/184/commits/43321d5c5aace948297596f136323f24c2e1835a)

### DIFF
--- a/src/server/plugins/engine/components/EmailAddressField.ts
+++ b/src/server/plugins/engine/components/EmailAddressField.ts
@@ -28,14 +28,7 @@ export class EmailAddressField extends FormComponent {
   }
 
   getViewModel(formData: FormData, errors?: FormSubmissionErrors) {
-    const schema = this.schema
     const viewModel = super.getViewModel(formData, errors)
-
-    if ('max' in schema && schema.max) {
-      viewModel.attributes = {
-        maxlength: schema.max
-      }
-    }
 
     viewModel.type = 'email'
     viewModel.autocomplete = 'email'
--- a/src/server/plugins/engine/views/partials/form.html
+++ b/src/server/plugins/engine/views/partials/form.html
@@ -1,7 +1,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList -%}
 
-<form method="post" enctype="multipart/form-data" autocomplete="on" novalidate>
+<form method="post" enctype="multipart/form-data" novalidate>
   <input type="hidden" name="crumb" value="{{crumb}}"/>
   {{ componentList(components) }}
 
--- a/src/server/plugins/engine/views/partials/form.html
+++ b/src/server/plugins/engine/views/partials/form.html
@@ -1,7 +1,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList -%}
 
-<form method="post" enctype="multipart/form-data" autocomplete="on">
+<form method="post" enctype="multipart/form-data" autocomplete="on" novalidate>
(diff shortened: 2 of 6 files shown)


src/server/plugins/engine/components/EmailAddressField.ts
```diff
@@ -28,15 +28,8 @@ export class EmailAddressField extends FormComponent {
   }
 
   getViewModel(formData: FormData, errors?: FormSubmissionErrors) {
-    const schema = this.schema
     const viewModel = super.getViewModel(formData, errors)
 
-    if ('max' in schema && schema.max) {
-      viewModel.attributes = {
-        maxlength: schema.max
```
colinrotherham, line 36: Can I double check where this change came from?

If we've removed features from fields we should do the same in the types/schema

https://github.com/DEFRA/forms-designer/blob/880d443c0a029be66a56da8193e2ee5279fd0737/model/src/components/types.ts#L23-L52

davidjamesstone: The property `max` is not getting removed from the field, it's still there and used to build the field's joi schema.

These removed lines just stop `maxlength` being passed as an attribute to the GDS Text Field njk macro, stopping:

`<input type="text" maxlength="10">`

alexluckett: Thanks for clarifying this Dave, it wasn't obvious to me either. I can see in the constructor that max is set on the componentSchema.

src/server/views/repeating-summary.html
```diff
@@ -11,7 +11,7 @@
       <div class="govuk-grid-column-two-thirds">
         <h1 class="govuk-heading-l">{{customText.separatePageTitle or pageTitle }}</h1>
         {{ govukSummaryList(details) }}
-        <form action="?view=summary"  method="post" enctype="multipart/form-data">
+        <form action="?view=summary"  method="post" enctype="multipart/form-data" novalidate>
```
colinrotherham, line 14: Tiny extra space from the previous change

```suggestion
        <form action="?view=summary" method="post" enctype="multipart/form-data" novalidate>
```

src/server/plugins/engine/views/partials/form.html
```diff
@@ -1,7 +1,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList -%}
 
-<form method="post" enctype="multipart/form-data" autocomplete="on">
+<form method="post" enctype="multipart/form-data" autocomplete="on" novalidate>
```
colinrotherham, line 4: Can we remove the default `autocomplete="on"`?

Not sure if affects anything but we should set [per-field `autocomplete` attributes](https://design-system.service.gov.uk/components/text-input/#use-the-autocomplete-attribute) instead

```suggestion
<form method="post" enctype="multipart/form-data" novalidate>
```

alexluckett (approved):
LGTM aside from the couple of minor things Colin spotted.

src/server/views/repeating-summary.html
davidjamesstone: Resolved with 4526d79ca3ba089732f7db668abb36d8bff5e2a2

(FWIW, this file, and none of the "repeating" stuff is actually used and can probably be deleted at some point)

src/server/plugins/engine/views/partials/form.html
davidjamesstone: Resolved with 4526d79ca3ba089732f7db668abb36d8bff5e2a2

PR #211, opened 2024-06-25
Add privacy policies

alexluckett:


src/server/plugins/router.ts
```diff
@@ -1,11 +1,23 @@
+import Boom from '@hapi/boom'
 import { type Request, type ResponseToolkit } from '@hapi/hapi'
 
+import { createLogger } from '../common/helpers/logging/logger.js'
+
```
colinrotherham, line 5:
```suggestion
import { createLogger } from '~/src/server/common/helpers/logging/logger.js'
```

davidjamesstone:
If it's your intention that `GET /help/privacy` (no slug) results in a 404, then 
`src/server/views/help/privacy.html` can be deleted. Otherwise, we'll need a guard something like:

```
if (!slug) {
  return h.view('help/privacy')
}
```

src/server/plugins/router.ts
```diff
@@ -32,9 +44,18 @@ export default {
...
-          path: '/help/privacy',
-          handler(_request: Request, h: ResponseToolkit) {
-            return h.view('help/privacy')
+          path: '/help/privacy/{slug?}',
+          handler(request: Request, h: ResponseToolkit) {
+            const { slug } = request.params
+
+            const privacyPolicy = privacyPolicies[slug as string]
+
+            if (!privacyPolicy) {
```
colinrotherham, line 53: The compiler was kicking off a bit because `slug` might be `undefined`

So if we do both checks it'll be happy

```suggestion
            if (!slug || !(slug in privacyPolicies)) {
```

src/server/plugins/router.ts
```diff
@@ -32,9 +44,18 @@ export default {
...
+              logger.error(`Privacy policy not found for slug ${slug}`)
+              throw Boom.notFound()
+            }
+
+            return h.redirect(privacyPolicy)
```
colinrotherham, line 58:
```suggestion
            return h.redirect(privacyPolicies[slug])
```

src/server/plugins/router.ts
```diff
@@ -32,9 +44,18 @@ export default {
         },
         {
           method: 'get',
-          path: '/help/privacy',
-          handler(_request: Request, h: ResponseToolkit) {
-            return h.view('help/privacy')
+          path: '/help/privacy/{slug?}',
+          handler(request: Request, h: ResponseToolkit) {
```
colinrotherham, line 48:
```suggestion
          handler(request: Request<{ Params: { slug?: string }}>, h: ResponseToolkit) {
```

src/server/plugins/router.ts
alexluckett: Done

davidjamesstone (approved):

src/server/plugins/router.ts
```diff
@@ -35,9 +46,21 @@ export default {
...
+            request: Request<{ Params: { slug: string } }>,
+            h: ResponseToolkit
+          ) {
+            const { slug } = request.params
+
+            const privacyPolicy = privacyPolicies[slug]
+
+            if (!privacyPolicy) {
+              logger.error(`Privacy policy not found for slug ${slug}`)
+              throw Boom.notFound()
```
davidjamesstone, line 60: FYI - if you prefer, in route handlers you can `return Boom.notFound()` rather than `throw`ing it.

davidjamesstone (approved):

PR #213, opened 2024-06-25
Make support pages work off slug to support multiple environments

alexluckett:


src/server/plugins/router.ts
```diff
@@ -1,7 +1,8 @@
 import { type Request, type ResponseToolkit } from '@hapi/hapi'
 
+import { getFormMetadata } from './engine/services/formsService.js'
```
colinrotherham, line 3: Lost the `~/src` for some reason and moved up

```suggestion
import { getFormMetadata } from '~/src/server/plugins/engine/services/formsService.js'
```

davidjamesstone (approved):

src/server/plugins/router.ts
```diff
@@ -16,7 +17,7 @@ export default {
           method: 'get',
           path: '/help/get-support/{slug?}',
           async handler(request: Request, h: ResponseToolkit) {
```
colinrotherham, line 19: Just to avoid the `as string` (because it might be `undefined`) you can add:

```suggestion
          async handler(request: Request<{ Params: { slug?: string }}>, h: ResponseToolkit) {
```

src/server/plugins/router.ts
```diff
@@ -16,7 +17,7 @@ export default {
...
-            const { slug } = request.params
+            const slug = request.params.slug as string
```
colinrotherham, line 20:
```suggestion
            const { slug } = request.params
```

src/server/plugins/router.ts
alexluckett: Whoops! Thanks

src/server/plugins/router.ts
alexluckett: I removed it and restored it, it'd be from the IDE when importing from the context menu. Fixed.

src/server/plugins/router.ts
alexluckett: Done

colinrotherham (approved):

PR #220, opened 2024-06-27
Change wording for email links in the phase banner

alexluckett:


src/server/views/layout.html
```diff
@@ -36,18 +36,19 @@
...
 {% block beforeContent %}
     {% if phaseTag %}
-        {% if feedbackLink %}
+        {% if 'mailto:' in feedbackLink %}
             {{ govukPhaseBanner({
                 tag: {
                     text: phaseTag | capitalize
                 },
-                html: 'This is a new service – your <a class="govuk-link" href=' + feedbackLink + '>feedback</a> will help us to improve it.'
+                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink + '">give your feedback by email</a>.'
```
colinrotherham, line 44: Can you double check URL encoding?

Mailto links will need `&` → `&amp;` etc

```console
FEEDBACK_LINK=mailto:hello@example.org?subject=Feedback&body=Hello
```

```suggestion
                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink | urlencode + '">give your feedback by email</a>.'
```

src/server/views/layout.html
```diff
@@ -36,18 +36,19 @@
...
             }) }}
         {% else %}
             {{ govukPhaseBanner({
                 tag: {
                     text: phaseTag | capitalize
-                }
+                },
+                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink + '">give your feedback (opens in new tab)</a>.'
```
colinrotherham, line 51: Probably the same here

We set up a nice [Nunjucks helper](https://github.com/DEFRA/forms-designer/blob/3d884bea6ad33cd42eb71dd83a9b19317ab3cb7e/designer/server/src/models/forms/form-lifecycle.js#L30-L33) for this in **forms-designer** which could be handy (another day)

alexluckett:
I've accidentally included two stray files relating to unicorn breeders. I'll fix these up when https://github.com/DEFRA/forms-runner/pull/228/files is merged in to this.

src/server/views/layout.html
alexluckett: Fixed in the following two commits.

src/server/views/layout.html
alexluckett: Fixed in the following two commits.

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -711,7 +729,7 @@ export class PageControllerBase {
     }
   }
 
-  feedbackUrlFromRequest(request: Request): string | void {
+  feedbackUrlFromRequest(request: Request): string | undefined {
```
colinrotherham, line 732: Cheers for this

Can we remove the duplicate method from **SummaryPageController** now?

It also extends from **PageControllerBase** but has its own version:

https://github.com/DEFRA/forms-runner/blob/86e70c53d5c2d93cbd59760ea0759b467f4cc651/src/server/plugins/engine/pageControllers/SummaryPageController.ts#L200

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -694,13 +697,28 @@ export class PageControllerBase {
     if (feedbackContextInfo) {
       viewModel.name = feedbackContextInfo.formTitle
     }
+
+    let feedbackLink: string | undefined = ''
+
     // setting the feedbackLink to undefined here for feedback forms prevents the feedback link from being shown
     if (this.def.feedback?.url) {
-      viewModel.feedbackLink = this.feedbackUrlFromRequest(request)
+      feedbackLink = this.feedbackUrlFromRequest(request)
```
colinrotherham, line 705: We've got the check for `this.def.feedback?.url` in **this.feedbackUrlFromRequest()** too

Shall we simplify all this?

```js
const { feedback } = this.def

let feedbackLink = this.feedbackUrlFromRequest(request)

if (!feedbackLink) {
  feedbackLink = feedback?.emailAddress
    ? `mailto:${feedback.emailAddress}`
    : config.feedbackLink
}
```

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -694,13 +697,28 @@ export class PageControllerBase {
...
+    } else if (this.def.feedback?.emailAddress) {
+      feedbackLink = `mailto:${this.def.feedback.emailAddress}`
+    } else {
+      feedbackLink = config.feedbackLink
     }
-    if (this.def.feedback?.emailAddress) {
-      viewModel.feedbackLink = `mailto:${this.def.feedback.emailAddress}`
+
+    if (feedbackLink?.startsWith('mailto:')) {
```
colinrotherham, line 712: Mind if we add tests into **PageControllerBase.test.ts** for this?

Would be ace to clearly see the escaping + error logging in action as currently missed these checks from the summary page view model ([see comment](https://github.com/DEFRA/forms-runner/pull/220/files#r1662041535))

Probably seen this, but we'll want a Jest env adding for `feedbackLink` so the unit tests run too

```console
The server config is invalid. "feedbackLink" is required
```

src/server/plugins/engine/pageControllers/PageControllerBase.ts
alexluckett: Aha, you have read my mind 😄 

I have already done this in a follow-up PR: https://github.com/DEFRA/forms-runner/pull/228

I went down a rabbit hole and it ended up getting too big so I spun it out separately.

src/server/plugins/engine/pageControllers/PageControllerBase.ts
alexluckett: Happy to. I'm ditching the `feedbackUrlFromRequest` method entirely as it was doing some stuff we don't need. In my clean up PR to follow https://github.com/DEFRA/forms-runner/pull/228 I'm just using `feedback.url` instead of the needlessly complex `this.feedbackUrlFromRequest(request)`.

src/server/plugins/engine/pageControllers/PageControllerBase.ts
alexluckett: Ah I missed this, thanks, I'll add it.

colinrotherham:
@alexluckett You alright sorting the conflicts out before I have another look?

alexluckett:
> @alexluckett You alright sorting the conflicts out before I have another look?

Conflicts sorted, just adding the tests next for page controller base.

colinrotherham:
Thanks @alexluckett

Looks like https://github.com/DEFRA/forms-runner/pull/220/commits/e4d21e3257722222809c30bb09edb9394192a10e suggests `FEEDBACK_LINK` is mandatory now, can we update the config?

For convict, you'll need to set `default: null` with types for a string value

```patch
  feedbackLink: {
    format: String,
-   default: '#', // Used in your phase banner. Can be a URL or more commonly mailto mailto:feedback@department.gov.uk
+   default: null,
    env: 'FEEDBACK_LINK'
- },
+ } as SchemaObj<string>,
```

I'll manually test the escaping unless tests are coming?

src/server/plugins/engine/pageControllers/PageControllerBase.ts
colinrotherham: I've done some testing and we should escape `http(s)://` URLs too

```js
    if (feedbackLink) {
      const isExternal = /^(https?|mailto):/.test(feedbackLink)
      const baseURL = !isExternal ? config.get('baseURL') : undefined // Need APP_BASE_URL in config

      try {
        feedbackLink = new URL(feedbackLink, baseURL).toString()
      } catch (err) {
        logger.error(err, `Failed to decode ${feedbackLink}`)
        feedbackLink = undefined
      }
    }
```

src/server/views/layout.html
```diff
@@ -36,18 +36,19 @@
...
-                html: 'This is a new service – your <a class="govuk-link" href=' + feedbackLink + '>feedback</a> will help us to improve it.'
+                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink | escape + '">give your feedback by email</a>.'
             }) }}
         {% else %}
             {{ govukPhaseBanner({
                 tag: {
                     text: phaseTag | capitalize
-                }
+                },
+                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink | escape + '">give your feedback (opens in new tab)</a>.'
```
colinrotherham, line 51: Need to add the "opens in new tab" behaviour ([see guidance](https://design-system.service.gov.uk/styles/links/#opening-links-in-a-new-tab))

```suggestion
                html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + feedbackLink | escape + '" target="_blank" rel="noopener noreferrer">give your feedback (opens in new tab)</a>.'
```

src/server/views/layout.html
```diff
@@ -36,18 +36,19 @@
 
 {% block beforeContent %}
     {% if phaseTag %}
-        {% if feedbackLink %}
+        {% if 'mailto:' in feedbackLink %}
```
colinrotherham, line 39: Ahh from testing, there's another case we've missed

When `feedbackLink` starts with `/` not `http(s)://` we'll need to drop the new tab behaviour

src/server/views/layout.html
colinrotherham: @alexluckett Escaping is working great, nice work 🙌

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -686,13 +690,26 @@ export class PageControllerBase {
     if (feedbackContextInfo) {
       viewModel.name = feedbackContextInfo.formTitle
     }
+
     // setting the feedbackLink to undefined here for feedback forms prevents the feedback link from being shown
-    if (this.def.feedback?.url) {
-      viewModel.feedbackLink = this.feedbackUrlFromRequest(request)
+    let feedbackLink: string | undefined = this.feedbackUrlFromRequest(request)
```
colinrotherham, line 695: Cheers for simplifying this

It can probably figure out the types from the return value if that helps

```suggestion
    let feedbackLink = this.feedbackUrlFromRequest(request)
```

colinrotherham (approved):

PR #228, opened 2024-06-28
Simplify feedback link code

alexluckett:
This simplifies how feedback links are generated. Previously, it passed context to the link containing information about the form, but we don't need this. Now we'll just use the link from the form definition as the priority, else fall back on config.

**Note:** previously pages relied on the global context with overrides performed within the view model if required. This was too fragile as it relied on the values not being set, but that instantly breaks with a view model like `SummaryViewModel` which returns `undefined` if the value is not set.

As a result I've been able to clear out a lot of code we don't need :-) 

Functional changes: 
- Phase banner uses the form's value if provided, else falls back on the server config
- Feedback link uses the form's value if provided, else falls back on the server config
- If the feedback link is malformed, fail fast and throw an exception. Prevents teams being ignorant to invalid forms rather than letting bad config creep into live.
- Remove support for dynamic feedback links with context passing. If feedback is required, we simply take them to an external link and don't bother with return URLs - nobody is asking for this feature, no point maintaining it.


alexluckett:
There's a few more pages the feedback link is used, so I'm pulling this into draft.

nayij:
can this be closed?

colinrotherham:
>can this be closed?

@nayij Maybe, but lots of unused feedback form code was removed in this PR

Would be great to split it out and get that bit merged

colinrotherham:
@alexluckett I've rebased this just to give the Sonar config a run

colinrotherham (approved):
Great to remove all this unused code

PR #236, opened 2024-07-01
Feature/404718 check answers branching

davidjamesstone:
[Missing anwsers summary links](https://github.com/DEFRA/forms-runner/pull/236/commits/f7f81be640a875c5848fa0714dde945fbe5a3c49)

Following the guidance in the [show missing information section](https://design-system.service.gov.uk/components/summary-list/)

[Add error summary to the summary page in the event of incomplete form data](https://github.com/DEFRA/forms-runner/pull/236/commits/5156e0b45b909413796d5a8829bc13b6f0ca0005)

### DIFF
--- a/src/server/views/summary.html
+++ b/src/server/views/summary.html
@@ -1,12 +1,21 @@
+{% extends 'layout.html' %}
+
 {% from "partials/summary-detail.html" import summaryDetail %}
 {% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
-{% extends 'layout.html' %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
 
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       {% if previewMode %}
         {% include "partials/preview-banner.html" %}
+      {% endif %}
+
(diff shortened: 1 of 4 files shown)


alexluckett:
I think this is a pretty solid addition. My only query really is how much further we can go without adding unneccessary complexity.

I'll chat with you tomorrow if you're feeling better.

alexluckett (approved):

src/server/views/summary.html
```diff
@@ -17,18 +26,16 @@ <h1 class="govuk-heading-l">
         {{ summaryDetail(detail) }}
       {% endfor %}
 
-      {% if not result.error  %}
-        <form method="post" autocomplete="off" novalidate>
-          <input type="hidden" name="crumb" value="{{crumb}}"/>
+      <form method="post" autocomplete="off" novalidate>
+        <input type="hidden" name="crumb" value="{{crumb}}"/>
```
colinrotherham, line 30: We only really do this for crumb, but one to replace the XHTML `/>` → `>` another day

```suggestion
        <input type="hidden" name="crumb" value="{{crumb}}">
```

colinrotherham:
Worth clarifying that this is a workaround in place of improved branching support?

PR #240, opened 2024-07-02
Move config into convict

colinrotherham:
This PR moves the **form-runner** config into [convict](https://www.npmjs.com/package/convict) like we do in **forms-manager** (and CDP templates)

```patch
const pluginPaths = [
- join(config.appDir, 'plugins/engine/views'),
- join(config.appDir, 'views')
+ join(config.get('appDir'), 'plugins/engine/views'),
+ join(config.get('appDir'), 'views')
]
```


alexluckett (approved):
Amazing work 🙂

PR #267, opened 2024-07-11
Remove build schema helpers

colinrotherham:
This PR removes the helpers `buildSchema()`, `buildStateSchema()` etc

These changes address the comment:

https://github.com/DEFRA/forms-runner/blob/79dfe9d92f57c4282ceecb5b8149fcd7a05547f8/src/server/plugins/engine/components/helpers.ts#L4-L8

### DIFF
--- a/src/server/plugins/engine/components/YesNoField.ts
+++ b/src/server/plugins/engine/components/YesNoField.ts
@@ -1,11 +1,7 @@
-import { type List, type YesNoFieldComponent } from '@defra/forms-model'
+import { type YesNoFieldComponent } from '@defra/forms-model'
 
 import { ListFormComponent } from '~/src/server/plugins/engine/components/ListFormComponent.js'
-import {
-  addClassOptionIfNone,
-  buildFormSchema,
-  buildStateSchema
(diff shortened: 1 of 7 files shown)


src/server/plugins/engine/components/YesNoField.ts
```diff
@@ -37,34 +33,15 @@ export class YesNoField extends ListFormComponent {
...
-  get items() {
-    return this.list?.items ?? []
-  }
-
-  get values() {
-    return [true, false]
-  }
-
   constructor(def: YesNoFieldComponent, model: FormModel) {
     super(def, model)
```
colinrotherham, line 37: @davidjamesstone I've let the Yes/No component inherit both schema from `ListFormComponent`

Do you know why it's casting `true` → `'true'` despite the boolean list type?

colinrotherham: The Yes/No list is already in the FormModel so this works:

```patch
  constructor(def: YesNoFieldComponent, model: FormModel) {
-   super(def, model)
+   super({ ...def, list: '__yesNo' }, model)
```

davidjamesstone (approved):

PR #272, opened 2024-07-12
Add custom error message

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/398579#18941354

- When a `customValidationMessage` is set for Email, Text, TextArea, Number and Telephone components, the message should be used for any errors for the component

### DIFF
--- a/src/server/plugins/engine/components/EmailAddressField.ts
+++ b/src/server/plugins/engine/components/EmailAddressField.ts
@@ -1,4 +1,5 @@
 import { type EmailAddressFieldComponent } from '@defra/forms-model'
+import joi from 'joi'
 
 import { TextField } from '~/src/server/plugins/engine/components/TextField.js'
 import { type FormModel } from '~/src/server/plugins/engine/models/index.js'
@@ -14,9 +15,25 @@ export class EmailAddressField extends TextField {
   constructor(def: EmailAddressFieldComponent, model: FormModel) {
     super(def, model)
 
-    const { schema, options } = def
+    const { schema, options, title } = def
 
-    this.formSchema = this.formSchema.email()
+    let formSchema = joi.string().label(title.toLowerCase()).email()
+
+    if (options.required === false) {
+      formSchema = formSchema.allow('').allow(null)
+    }
+
+    if (options.customValidationMessage) {
+      const message = options.customValidationMessage
+
+      formSchema = formSchema.messages({
+        'any.required': message,
+        'string.empty': message,
+        'string.email': message
+      })
+    }
+
+    this.formSchema = formSchema
     this.options = options
     this.schema = schema
   }
--- a/src/server/plugins/engine/components/EmailAddressField.ts
+++ b/src/server/plugins/engine/components/EmailAddressField.ts
@@ -17,7 +17,7 @@ export class EmailAddressField extends FormComponent {
 
     const { schema, options, title } = def
 
-    let formSchema = joi.string().label(title.toLowerCase()).email()
+    let formSchema = joi.string().trim().label(title.toLowerCase()).email()
 
     if (options.required === false) {
       formSchema = formSchema.allow('').allow(null)
--- a/src/server/plugins/engine/components/EmailAddressField.ts
+++ b/src/server/plugins/engine/components/EmailAddressField.ts
@@ -1,14 +1,14 @@
 import { type EmailAddressFieldComponent } from '@defra/forms-model'
 import joi from 'joi'
 
-import { TextField } from '~/src/server/plugins/engine/components/TextField.js'
+import { FormComponent } from '~/src/server/plugins/engine/components/FormComponent.js'
 import { type FormModel } from '~/src/server/plugins/engine/models/index.js'
 import {
   type FormPayload,
   type FormSubmissionErrors
 } from '~/src/server/plugins/engine/types.js'
 
-export class EmailAddressField extends TextField {
+export class EmailAddressField extends FormComponent {
   declare options: EmailAddressFieldComponent['options']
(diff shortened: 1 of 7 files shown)


src/server/plugins/engine/components/TelephoneNumberField.ts
```diff
@@ -9,8 +9,8 @@ import {
   type FormSubmissionErrors
 } from '~/src/server/plugins/engine/types.js'
 
-const PATTERN = /^[0-9\\\s+()-]*$/
-const DEFAULT_MESSAGE = 'Enter a telephone number in the correct format'
+const PATTERN =
+  /\s*(([+](\s?\d)([-\s]?\d)|0)?(\s?\d)([-\s]?\d){9}|[(](\s?\d)([-\s]?\d)+\s*[)]([-\s]?\d)+)\s*/
```
colinrotherham, line 13: @davidjamesstone Are phone numbers without spaces still valid?

We should differentiate between:

1. Values users can enter as they prefer via `formSchema`
2. Values that are formatted for state via `stateSchema`

From the [guidance for **Phone numbers**](https://design-system.service.gov.uk/patterns/phone-numbers/) it says:

>**Allow different formats**
>
>Let users enter phone numbers in whatever format is familiar to them. Allow for additional spaces, hyphens, dashes and brackets, and be able to accommodate country and area codes.

src/server/plugins/engine/components/TextField.ts
```diff
@@ -51,8 +54,15 @@ export class TextField extends FormComponent {
...
+      const message = options.customValidationMessage
+
       formSchema = formSchema.messages({
-        any: options.customValidationMessage
+        'any.required': message,
+        'string.empty': message,
+        'string.max': message,
+        'string.min': message,
+        'string.length': message,
+        'string.pattern.base': message
```
colinrotherham, line 65: Can I check we definitely want `customValidationMessage` to override all of these?

For example, a custom message for invalid formatting will also show for min/max length

test/cases/server/plugins/engine/TelephoneNumberField.test.ts
```diff
@@ -1,12 +1,13 @@
 import { ComponentType, type ComponentDef } from '@defra/forms-model'
 
 import { TelephoneNumberField } from '~/src/server/plugins/engine/components/index.js'
+import { validationOptions as opts } from '~/src/server/plugins/engine/pageControllers/validationOptions.js'
```
colinrotherham, line 4: I like this, I'll use `opts` in all my new tests too

src/server/plugins/engine/components/TelephoneNumberField.ts
davidjamesstone: The new regex is liberal and allows users to enter in whatever format they like (see PR description for examples), including brackets and dashes etc.

It just improves it to not match on things like this that were in the tests:

``` 
expect(schema.validate('+111-111-11').error).toBeUndefined()
expect(schema.validate('+111 111 11').error).toBeUndefined()
expect(schema.validate('+11111111').error).toBeUndefined()
```

colinrotherham: > @davidjamesstone Are phone numbers without spaces still valid?

It looks like it, but can definitely still place spaces where you like?

But from the guidance extract above, maybe this style needs some work?

> (+44) 07777 111222
> +44 (0)7777 111222

davidjamesstone: They're both acceptable. Do you want them adding to the tests?

davidjamesstone: As are variants of the same e.g.

(+44) 07777 111 222
(+44) 07777 111-222
(+44) 07777-111222
(+44) 07777-111-222

etc...

colinrotherham: Might as well, no harm adding them. Thanks for confirming

src/server/plugins/engine/components/EmailAddressField.ts
```diff
@@ -14,9 +15,25 @@ export class EmailAddressField extends TextField {
...
 
-    const { schema, options } = def
+    const { schema, options, title } = def
 
-    this.formSchema = this.formSchema.email()
+    let formSchema = joi.string().label(title.toLowerCase()).email()
+
+    if (options.required === false) {
+      formSchema = formSchema.allow('').allow(null)
+    }
```
colinrotherham, line 24: Presume we should just tag onto `this.formSchema` from the extended text field?

All of these bits are set in `super(def, model)` although `.messages()` are different

```suggestion
```

src/server/plugins/engine/components/MultilineTextField.ts
```diff
@@ -33,41 +33,68 @@ export class MultilineTextField extends FormComponent {
...
-    let formSchema = Joi.string().label(title.toLowerCase())
+    let formSchema = Joi.string().trim().label(title.toLowerCase())
 
     if (isRequired) {
       formSchema = formSchema.required()
     } else {
       formSchema = formSchema.allow('').allow(null)
     }
 
-    formSchema = formSchema.ruleset
```
colinrotherham, line 44: What does removing this do?

colinrotherham (approved):
Approved! Can we flag the regex issue for further review?

davidjamesstone:
Remove this regex for review later.

`/\s*(([+](\s?\d)([-\s]?\d)|0)?(\s?\d)([-\s]?\d){9}|[(](\s?\d)([-\s]?\d)+\s*[)]([-\s]?\d)+)\s*/`

PR #284, opened 2024-07-18
Prevent date field showing error messages from other fields

colinrotherham:
This PR fixes [bug #415548](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/415548)

Two parts are resolved in the same related bug:

1. Date fields (incorrectly) show the first field's error message
2. Month & year fields (incorrectly) omit the error message and red border

# Part 1: Date fields
Only shows the first field's error message

Incorrect error message

# Part 1: Month & year fields
No error message or red left border is displayed

## Before

Before

## After

After

### DIFF
--- a/src/server/plugins/engine/components/DatePartsField.ts
+++ b/src/server/plugins/engine/components/DatePartsField.ts
@@ -131,10 +131,12 @@ export class DatePartsField extends FormComponent {
   }
 
   getViewModel(payload: FormPayload, errors?: FormSubmissionErrors) {
+    const { children, name } = this
+
     const viewModel = super.getViewModel(payload, errors)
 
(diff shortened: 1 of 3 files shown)


src/server/plugins/engine/components/DatePartsField.ts
```diff
@@ -151,8 +153,14 @@ export class DatePartsField extends FormComponent {
       }
     })
 
-    const firstError = errors?.errorList[0]
```
colinrotherham, line 154: This is the bug, it assumed the errors are for the current component only

(Errors should be filtered by name instead)

alexluckett (approved):

PR #285, opened 2024-07-18
Ensure request logger only logs errors in development

colinrotherham:
This PR applies change already made in **forms-designer** via https://github.com/DEFRA/forms-designer/commit/c2ba33671e2bce7951e9929744ff99a028346cc8

To reduce noise, only errors will be logged via `npm run dev`

But when necessary `LOG_LEVEL` can be set using:

```console
LOG_LEVEL=info npm run dev
```

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -159,7 +159,7 @@ export const config = convict({
   logLevel: {
     doc: 'Logging level',
     format: ['fatal', 'error', 'warn', 'info', 'debug', 'trace', 'silent'],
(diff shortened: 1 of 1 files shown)


alexluckett (approved):
Approving for consistency with forms-designer.

@davidjamesstone @KiranVarma02 FYI, you'll have to set LOG_LEVEL in your `.env` file if you want the existing behaviour.

We log at INFO by default in our deployed environment, this won't reduce logs for running applications.

PR #317, opened 2024-08-05
Transform dates to yyyy-MM-dd before evaluating conditions

alexluckett:
Date conditions were previously evaluated as datetime strings (ISO 8601 format). This worked for greater/less than conditions, but failed for quality conditions as `2024-08-01` would never equal `2024-08-01T00:00:00.000Z`. This code transforms the value to the expected format prior to condition evaluation.

@davidjamesstone FYI, implemented as discussed on Friday.

### DIFF
--- a/src/server/plugins/engine/components/DatePartsField.test.ts
+++ b/src/server/plugins/engine/components/DatePartsField.test.ts
@@ -1,9 +1,25 @@
 import { ComponentType, type ComponentDef } from '@defra/forms-model'
 
 import { DatePartsField } from '~/src/server/plugins/engine/components/DatePartsField.js'
-import { type FormModel } from '~/src/server/plugins/engine/models/FormModel.js'
+import { FormModel } from '~/src/server/plugins/engine/models/FormModel.js'
+import { type FormDefinition } from '~/src/server/plugins/engine/services/formsService.js'
+
+
 
 describe('Date parts field', () => {
+  let formModel: FormModel
+
+  beforeEach(() => {
+    const definition: FormDefinition = {
+      pages: [],
+      lists: [],
+      sections: [],
+      conditions: []
+    }
+
+    formModel = new FormModel(definition, { basePath: '' })
+  })
+
   test('Should construct appropriate children when required', () => {
     const def: ComponentDef = {
       name: 'myComponent',
@@ -13,7 +29,7 @@ describe('Date parts field', () => {
       schema: {}
     }
 
-    const underTest = new DatePartsField(def, {} as FormModel) // FormModel param not required for testing
+    const underTest = new DatePartsField(def, formModel)
     const returned = underTest.getViewModel({})
 
     expect(returned.fieldset).toEqual({
@@ -38,7 +54,7 @@ describe('Date parts field', () => {
       schema: {}
     }
 
-    const underTest = new DatePartsField(def, {} as FormModel) // FormModel param not required for testing
+    const underTest = new DatePartsField(def, formModel)
     const returned = underTest.getViewModel({})
 
     expect(returned.fieldset).toEqual({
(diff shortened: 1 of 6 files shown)


src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -1,4 +1,10 @@
-import { ComponentType, type ComponentDef } from '@defra/forms-model'
+import {
+  ComponentType,
+  type DatePartsFieldFieldComponent,
+  type ComponentDef
+} from '@defra/forms-model'
+
+import { type FormModel } from '../models/FormModel.js'
```
colinrotherham, line 7: Mind doing the usual `~` here?

src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -12,7 +18,7 @@ describe('Date parts field', () => {
       schema: {}
     }
 
-    const underTest = new DatePartsField(def, {})
+    const underTest = new DatePartsField(def, {} as FormModel) // FormModel param not required for testing
```
colinrotherham, line 21: Can you leave out the `as` casting please? I've slowly been fixing things like this (pretending something is a different type)

src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -73,11 +79,33 @@ describe('Date parts field', () => {
...
     expect(underTest.getViewModel({}).errorMessage).toBeUndefined()
   })
+  test('Condition evaluation used yyyy-MM-dd format', () => {
+    const datePartsFieldComponent = {
+      title: 'Example checkboxes',
+      name: 'myComponent',
+      type: ComponentType.DatePartsField,
+      options: {},
+      schema: {}
+    } satisfies DatePartsFieldFieldComponent
```
colinrotherham, line 95: Did linting fail with the general purpose `satisfies ComponentDef` here?

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -17,6 +17,8 @@ import { format, parseISO } from 'date-fns'
 import joi from 'joi'
 import { type ValidationResult, type ObjectSchema } from 'joi'
 
+import { DatePartsField } from '../components/DatePartsField.js'
```
colinrotherham, line 20: Another one for `~` thanks @alexluckett

src/server/plugins/engine/components/DatePartsField.test.ts
alexluckett: Nah, I just referenced another test file and didn't notice the ComponentDef earlier in this one. I've switched to ComponentDef for consistency within the file.

colinrotherham: Yeah, either works for me, can stick to one only most of the time. Thanks

src/server/plugins/engine/components/DatePartsField.test.ts
colinrotherham: Ideally upgrade to one of the newer test files where we instantiate `new FormModel()`

E.g. Here's the one for email component:

https://github.com/DEFRA/forms-runner/blob/31e4713d105065272db5b15535d878332cc66e61/src/server/plugins/engine/components/EmailAddressField.test.ts#L19-L25

With a 100% valid `new EmailAddressField(def, formModel)` for each test:

https://github.com/DEFRA/forms-runner/blob/31e4713d105065272db5b15535d878332cc66e61/src/server/plugins/engine/components/EmailAddressField.test.ts#L32-L43

alexluckett: Good timing, I just committed exactly that :). Thanks, done!

src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -1,8 +1,25 @@
...
+
+  beforeEach(() => {
+    const definition: FormDefinition = {
+      pages: [],
+      lists: [],
+      sections: [],
+      conditions: []
+    }
+
+    formModel = new FormModel(definition, { basePath: '' })
```
colinrotherham, line 20: Mind if keep `basePath: 'test'` I've been using in all the new tests?

Think of it as our test slug

```suggestion
    formModel = new FormModel(definition, {
      basePath: 'test'
    })
```

colinrotherham (approved):

colinrotherham:
@alexluckett Can you run `npm run format` too please?

PR #337, opened 2024-08-14
#412322 Request persistence for files in form submission upon upload

alexluckett:
When submitting a form, call `http://forms-submission-api/files/persist` to persist them for 30 days before sending an email.

Paired with https://github.com/DEFRA/forms-submission-api/pull/37.


src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -225,12 +230,54 @@ export class SummaryPageController extends PageController {
...
+  // Find out our file upload components
+  const fileUploadComponentNames = model.pages.flatMap((page) =>
+    page.components.formItems
+      .filter((item) => item.type === ComponentType.FileUploadField)
+      .map((item) => item.name)
+  )
+
+  // Pull the state for each file upload component
+  const fileUploadStates = fileUploadComponentNames.map(
+    (name) => state[name] as FileState[]
```
davidjamesstone, line 258: `state[name]` will be null for optional file upload components - might be worth a guard

davidjamesstone (approved):

src/server/plugins/engine/pageControllers/SummaryPageController.ts
colinrotherham: Mind if we avoid `as` unless we're absolutely stuck and/or can't teach the compiler the types itself?

Arrays are mutable so TypeScript can intentionally not trust types when mapping

Adding a quick `Readonly<>` or `as const` can often fix this sort of thing

davidjamesstone: Also need to consider the case when the component is nested under a section.

Something like `const pageState = this.section ? state[this.section.name] : state` or use `page.getFormDataFromState{state}`

image

src/server/plugins/engine/services/formSubmissionService.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+export async function persistFiles(files, persistedRetrievalKey) {
+  const postJsonByType = /** @type {typeof postJson} */ (postJson)
+
+  const payload = {
+    files,
+    persistedRetrievalKey
+  }
+
+  logger.info(`Completed persistence request`)
```
alexluckett, line 21: This shouldn't be here

davidjamesstone: Resolved with ffac4d1de8de2586f8473d8043ed987bdf09605f

PR #364, opened 2024-08-30
Wrap user submitted answers in literal MD codeblock

davidjamesstone:


src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -379,6 +380,12 @@ export function getPersonalisation(
   }
 }
 
+function literal(str: string) {
+  return `\`\`\`
+${str}
+\`\`\``
+}
```
colinrotherham, line 387: Do we need a trailing new line and/or Markdown `text` flavour?

```suggestion
function literal(str: string) {
  return `\`\`\`text\n${str}\n\`\`\``
}
```

colinrotherham (approved):

PR #375, opened 2024-09-03
Feature/max future past days tests

davidjamesstone:


src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -113,6 +118,178 @@ describe('Date parts field', () => {
...
+      )
+    })
+  })
+
+  describe('State validation', () => {
+    const now = new Date()
+    const _1DayInPast = addDays(now, -1)
+    const _2DaysInPast = addDays(now, -2)
+    const _1DayInFuture = addDays(now, 1)
+    const _2DaysInFuture = addDays(now, 2)
```
colinrotherham, line 215: Can we put these inline?

Probably just as clear without the new naming convention

I do still prefer [`luxon`](https://moment.github.io/luxon/#/) `DateTime.now().minus({ days: 2 })` 😆

```suggestion
    const now = new Date()
```

colinrotherham (approved):

src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -113,6 +118,178 @@ describe('Date parts field', () => {
 
     expect(conditonEvaluationStateValue).toBe('2024-12-31')
   })
+
+  describe('Form validation', () => {
```
colinrotherham, line 122: Should probably match the other tests

```suggestion
  describe('Validation', () => {
```

src/server/plugins/engine/components/DatePartsField.test.ts
```diff
@@ -113,6 +118,178 @@ describe('Date parts field', () => {
...
+          const { formSchema } = component.children
+
+          const result = formSchema.validate(input, opts)
+          expect(result).toEqual(output)
+        }
+      )
+    })
+  })
+
+  describe('State validation', () => {
```
colinrotherham, line 210: Should probably match the other tests

```suggestion
  describe('State', () => {
```

PR #380, opened 2024-09-06
Use privacy notice from form metadata

davidjamesstone:


src/server/plugins/router.ts
```diff
@@ -60,17 +37,18 @@ export default {
...
+        async handler(request, h) {
           const { slug } = request.params
 
-          const privacyPolicy = privacyPolicies[slug]
+          const form = await getFormMetadata(slug)
+          const { privacyNoticeUrl } = form
 
-          if (!privacyPolicy) {
-            logger.error(`Privacy policy not found for slug ${slug}`)
+          if (!privacyNoticeUrl) {
```
alexluckett, line 46: We'll need to manually update the forms in prod with the privacy policies from this deleted code. I suggest we do it this week so it's fresh in our minds and we don't forget.

alexluckett (approved):

src/server/plugins/router.ts
davidjamesstone: Yes I was thinking the same thing. I've added them to [the ticket](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/290022) for reference

PR #402, opened 2024-09-19
feat: [416317] Manage file upload for js users

KiranVarma02:


alexluckett:
What's happening with this PR? @colinrotherham  are you making us of it now you've taken over the story?

colinrotherham:
@alexluckett Yeah think so. Exploring if it's easier just to return (and replace) the entire HTML fragment

colinrotherham:
I'm going to close this PR

We'll be removing the reload/refresh link so won't need a solution (like this) to automatically poll for changes

PR #409, opened 2024-09-24
bug: [384242] Add hint to address page

KiranVarma02:


src/server/plugins/engine/components/UkAddressField.ts
```diff
@@ -186,6 +179,13 @@ export class UkAddressField extends FormComponent {
       }
     }
 
+    if (hint) {
+      hint.id ??= `${name}-hint`
```
colinrotherham, line 183: We need to set the hint ID so it matches the fieldset `aria-describedby`

src/server/plugins/engine/components/UkAddressField.ts
```diff
@@ -186,6 +179,13 @@ export class UkAddressField extends FormComponent {
...
+      fieldset.attributes ??= {
+        'aria-describedby': hint.id
```
colinrotherham, line 185: Here we link the fieldset (by hint ID) so screen readers describe it using the hint text

src/server/plugins/engine/views/components/ukaddressfield.html
```diff
@@ -1,12 +1,25 @@
...
+        text: component.model.hint.text
+      } if fieldset else component.model.hint) }}
+    {% endset %}
+
+    {% set addressFieldHtml = addressHintHtml + addressFieldHtml %}
+  {% endif %}
+
   {{ govukFieldset({
     legend: fieldset.legend,
+    attributes: fieldset.attributes,
```
colinrotherham, line 22: Pass our new `aria-describedby` attribute to the `<fieldset>`

colinrotherham (approved):

PR #430, opened 2024-10-03
Bugfix #446245: Conditions with sections

alexluckett:


alexluckett:
With Jignesh for review.

src/server/plugins/engine/models/SummaryViewModel.ts
```diff
@@ -171,33 +168,13 @@ export class SummaryViewModel {
   }
 }
 
-function gatherRepeatPages(state: FormSubmissionState) {
```
colinrotherham, line 174: 🎉

src/server/plugins/engine/models/FormModel.ts
```diff
@@ -148,30 +148,10 @@ export class FormModel {
...
-            [section.name]: sectionSchema
-          })
-        } else {
-          sectionPages.forEach((sectionPage) => {
-            schema = schema.concat(sectionPage.stateSchema)
-          })
-        }
-      }
+    relevantPages.forEach((page) => {
+      schema.concat(page.stateSchema)
```
colinrotherham, line 154: Can we put the ~`const`~ `let` back above and re-assign the schema every time it changes?

Think we should have code coverage for this (because everything still passed) 😱

```suggestion
      schema = schema.concat(page.stateSchema)
```

alexluckett: Ahh!! Good spot, thank you.

colinrotherham (approved):
Looks loads better 🎉

Would be nice to start adding model tests but not a blocker

alexluckett:
Waiting for Dave to merge #443 before merging.

davidjamesstone (approved):

PR #438, opened 2024-10-04
feat: [419250] Enable sending email from preview links

KiranVarma02:


src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -342,6 +343,10 @@ export function getPersonalisation(
     model
   )
 
+  const formName = isPreview
+    ? `TEST FORM SUBMISSION: ${model.name}`
+    : `Form received: ${model.name}`
```
alexluckett, line 348: This is not a form name any more but a subject line instead. Rename to `subject`?

src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -302,8 +298,12 @@ async function sendEmail(
 ) {
   request.logger.info(['submit', 'email'], 'Preparing email')
 
+  const { path } = request
+
+  const isPreview = path.toLowerCase().startsWith(PREVIEW_PATH_PREFIX)
```
alexluckett, line 303: We're duplicating this logic. Better to pull this out into a shared function so they don't drift.

KiranVarma02: ahh ok, i didn't see its used in `plugin.ts`. I will refactor that

KiranVarma02: done

src/server/plugins/engine/pageControllers/SummaryPageController.ts
KiranVarma02: done

alexluckett (approved):

src/server/plugins/engine/helpers.ts
```diff
@@ -77,3 +79,7 @@ export const filesize = (bytes: number) => {
 
   return Math.max(bytes, 0.1).toFixed(1) + byteUnits[i]
 }
+
+export const checkIfPreview = (path: string) => {
```
colinrotherham, line 83: Mind if we use a named function for stack traces and add some tests?

```suggestion
export function hasPreviewPath(path: string) {
```

src/server/plugins/engine/pageControllers/SummaryPageController.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import { format } from 'date-fns'
+
+import { getPersonalisation } from './SummaryPageController.js'
+
+import {
+  FormModel,
+  SummaryViewModel
+} from '~/src/server/plugins/engine/models/index.js'
+
+describe('getPersonalisation', () => {
```
colinrotherham, line 16: Think the old description has been left in?

Or if not, let's move and rename this describe block

```suggestion
describe('SummaryPageController', () => {
  describe('getPersonalisation', () => {
```

src/server/plugins/engine/pageControllers/SummaryPageController.test.ts
```diff
@@ -0,0 +1,69 @@
+import {
+  ComponentType,
+  type ComponentDef,
+  type FormDefinition
+} from '@defra/forms-model'
+import { type Request } from '@hapi/hapi'
+import { format } from 'date-fns'
+
+import { getPersonalisation } from './SummaryPageController.js'
```
colinrotherham, line 9: Another path for `~/src` please (sorry)

src/server/plugins/engine/pageControllers/SummaryPageController.ts
colinrotherham: Just flagging this because it still says `formName`?

colinrotherham (changes requested):

colinrotherham (approved):

PR #444, opened 2024-10-08
Remove schema `.optional()` when value is always submitted

colinrotherham:
This PR is a supporting change for [bug #428455](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/428455) to prevent double error messages

Only yes/no, radios and checkboxes are now `.optional()` since their values are not always POST'd

Other fields now default to empty strings `''` where appropriate

### DIFF
--- a/src/server/plugins/engine/components/AutocompleteField.test.ts
+++ b/src/server/plugins/engine/components/AutocompleteField.test.ts
@@ -94,9 +94,9 @@ describe.each([
 
         const { formSchema } = componentOptional
 
-        expect(formSchema.describe().flags).toEqual(
+        expect(formSchema.describe()).toEqual(
           expect.objectContaining({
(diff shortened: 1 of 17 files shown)


src/server/plugins/engine/components/AutocompleteField.test.ts
```diff
@@ -132,7 +132,7 @@ describe.each([
 
         expect(result.error).toEqual(
           expect.objectContaining({
-            message: expect.stringContaining(`Select ${label}`)
+            message: `Select ${label}`
```
colinrotherham, line 135: See PR https://github.com/DEFRA/forms-runner/pull/445 to fix "Select…" → "Enter…" for Autocomplete

davidjamesstone (approved):

PR #456, opened 2024-10-10
Add missing CSRF token to new forms

davidjamesstone:


src/server/plugins/engine/views/repeat-item-delete.html
```diff
@@ -18,6 +18,7 @@
       {% include "partials/heading.html" %}
 
       <form method="post" novalidate>
+        <input type="hidden" name="crumb" value="{{crumb}}"/>
```
colinrotherham, line 21: Doesn't have to be now (the others do it too) but it's probably the only XHML we still have

```suggestion
        <input type="hidden" name="crumb" value="{{crumb}}">
```

colinrotherham (approved):

PR #470, opened 2024-10-15
Review and check types for form/state values

colinrotherham:
This PR fixes all remaining ESLint errors

We should review the suggested guards in this PR to plan any necessary session storage changes

### Checking nested session values
E.g. Ensuring array values are for file uploads and not checkboxes

```patch
  getFormValueFromState(state: FormSubmissionState) {
    const { name } = this
+   const values = state[name]

-   if (Array.isArray(state[name])) {
-     return state[name]
+   if (!Array.isArray(values)) {
+     return []
    }
+
+   return values.filter(
+     (value): value is FileState =>
+       typeof value === 'object' && 'uploadId' in value
+   )
  }
```

### Reducing nested session values

By preferring flat fields in session state we can use existing `ComponentCollection()` methods to read and write between payload, state and display values out the box (without needing to convert to/from objects) 

#### Flat fields (used by Month/Year)

```json
"name__field1": "XXX",
"name__field2": "XXX",
"name__field3": "XXX"
```

#### Nested objects (used by UK Address)

```json
"name": {
  "field1": "XXX",
  "field2": "XXX",
  "field3": "XXX"
}
```

## Before
Output from `npm run lint:js`

```console
✖ 38 problems (3 errors, 35 warnings)
```

## After
Output from `npm run lint:js`

```console
✖ 6 problems (0 errors, 6 warnings)
```


src/server/plugins/engine/components/CheckboxesField.ts
```diff
@@ -43,43 +43,42 @@ export class CheckboxesField extends SelectionControlField {
...
     let { items, value } = viewModel
 
-    let payloadItems: string[] | number[] = []
-
     // Allow strings (temporarily) via controller renderWithErrors()
-    payloadItems = Array.isArray(value) ? value : payloadItems
-    payloadItems = typeof value === 'string' ? [value] : payloadItems
+    const payloadItems = [value]
+      .flat()
+      .filter((value) => typeof value === 'string' || typeof value === 'number')
```
colinrotherham, line 75: We can tweak this bit should we keep https://github.com/DEFRA/forms-runner/pull/296

By sanitising the payload we'll only have array values so can drop the `[value].flat()`

colinrotherham:
@alexluckett Mind taking another look at this now we're at 90% test coverage?

I'm happy to split out the "flat state" from the ESLint + type fixes if it helps

src/server/plugins/engine/models/SummaryViewModel.ts
```diff
@@ -227,7 +230,7 @@ function Item(
     path: page.path,
     label: component.title,
     value: component.getDisplayStringFromState(state),
-    rawValue: state[component.name],
+    rawValue: component.getFormValueFromState(state),
```
alexluckett, line 233: ❤️

alexluckett (approved):

PR #473, opened 2024-10-16
Feature/300762 submit

davidjamesstone:


test/form/journey-basic.test.js
```diff
@@ -13,6 +14,7 @@ const testDir = dirname(fileURLToPath(import.meta.url))
 
 jest.mock('~/src/server/utils/notify.ts')
 jest.mock('~/src/server/plugins/engine/services/formsService.js')
+jest.mock('~/src/server/plugins/engine/services/formSubmissionService.js')
```
colinrotherham, line 17: 🎉

src/server/plugins/engine/models/SummaryViewModel.ts
```diff
@@ -190,6 +166,51 @@ export class SummaryViewModel {
...
+  model: FormModel,
+  state: FormSubmissionState,
+  items: DetailItem[]
+) {
+  const { options } = page.repeat
+  const { name, title } = options
+  const repeatSummaryPath = page.getSummaryPath(request)
+  const path = `/${model.basePath}${page.path}`
+  const rawValue = state[options.name]
+  const isInitialised = Array.isArray(rawValue)
```
colinrotherham, line 181: Wondering if it could ever be an empty array?

If we could make your class method public it's handled in `getListFromState()`

```suggestion
  const rawValue = page.getListFromState(state)
  const isInitialised = !!rawValue.length
```

src/server/plugins/engine/models/SummaryViewModel.ts
```diff
@@ -190,6 +166,51 @@ export class SummaryViewModel {
...
+    returnUrl: redirectUrl(request, `/${model.basePath}/summary`)
+  })
+
+  const subItems: DetailItem[][] = []
+
+  if (isInitialised) {
+    rawValue.forEach((itemState) => {
+      const sub: DetailItem[] = []
+      for (const component of page.components.formItems) {
+        const item = Item(request, component, itemState, page, model)
```
colinrotherham, line 193: With us drilling down from `state` into `itemState`

We've still got `Item()` typed to accept `FormSubmissionState`

Do we need safeguards to make sure this never blows up?

```js
rawValue: state[component.name],
```

colinrotherham (commented):
Thanks @davidjamesstone

Are you alright fixing the `[]` type?

I think it's probably going to be:

```patch
  export interface DetailItemRepeat extends DetailItemBase {
-   rawValue: [] | null
+   rawValue: FormData[] | null
  }
```

davidjamesstone:
> Thanks @davidjamesstone
> 
> Are you alright fixing the `[]` type?
> 
> I think it's probably going to be:
> 
> ```diff
>   export interface DetailItemRepeat extends DetailItemBase {
> -   rawValue: [] | null
> +   rawValue: FormData[] | null
>   }
> ```

Resolved with 35749bfba3f8433f8939bc707cf8ef275087f7b9

src/server/plugins/engine/models/SummaryViewModel.ts
davidjamesstone: If someone were to land at the repeater summary page directly without adding any items then yes, it could be empty. If they arrived there as a new user/uninitialised session then it could also be undefined (isInitialised would be false)

src/server/plugins/engine/models/SummaryViewModel.ts
davidjamesstone: I'd argue that would be exceptional circumstances. I'm happy to do something like:

`const item = Item(request, component, (itemState ?? {}), page, model)`

but I just can't see how that could happen and in the event of an undefined/null item getting into the repeater array, it would probably warrant an exception.

colinrotherham: Only flagging it because it's no longer `FormSubmissionState`

E.g. It definitely doesn't have `progress: []` in there

1. Payload `FormPayload` (including `crumb`) extends `FormData`
2. Redis state `FormSubmissionState` extends `FormState` perhaps

We're not necessarily fixing issues here, we're putting the safeguards in place for future work

Can leave it for now

src/server/plugins/engine/models/SummaryViewModel.ts
colinrotherham: Thanks for clarifying. There's no `null` ever?

Would you mind if we used `page.getListFromState(state)` to give us a single "get a list" approach?

src/server/plugins/engine/models/SummaryViewModel.ts
```diff
@@ -190,6 +166,51 @@ export class SummaryViewModel {
...
+  }
+
+  items.push({
+    name,
+    path: page.path,
+    label: isInitialised ? `${title}s added` : title,
+    value: `You added ${value} ${title}${value === '1' ? '' : 's'}`,
+    rawValue,
+    url,
+    pageId: path,
```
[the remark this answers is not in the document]
alexluckett: +1 if we're not using it, great to drop it

src/server/plugins/engine/models/SummaryViewModel.ts
alexluckett: I think that'd be good. I'm personally not a fan of accessing the raw state directly where it can be avoided, moving the access into a single function where the logic can be reused would be handy.

Not a blocker though, we do this all over the place as it is.

colinrotherham: Plus it helps me out for Monday, where I've already set up a single method for `rawValue` 😆

davidjamesstone: > Thanks for clarifying. There's no `null` ever?

No definitely no `null`. We've been asked not to support optional repeaters (`options.min` should be positive), I'd like that to change and if it does, we'll probably choose `[]` not `null` as a marker in state to flag that the repeater page has been visited, but 0 items were added.

src/server/plugins/engine/models/SummaryViewModel.ts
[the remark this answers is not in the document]
davidjamesstone: Resolved with 750edac1b911881e2fa6658076f2f272d54d0cd6

alexluckett (approved):
Comments pending but not blockers IMO, approved

src/server/plugins/engine/models/SummaryViewModel.ts
davidjamesstone: Resolved with a6e0ae3b2e7ef65b7560bb0b5a39d0a9b49dc016

davidjamesstone:
> Comments pending but not blockers IMO, approved

I think that's all comments addressed - I'll merge shortly

PR #475, opened 2024-10-16
bug: [418692] Optional tag missing for single component when set to optional

KiranVarma02:
https://dev.azure.com/defragovuk/DEFRA-CDP/_boards/board/t/DXT%20Team/Stories?System.IterationPath=%40currentIteration&workitem=428692

### DIFF
--- a/src/server/plugins/engine/pageControllers/PageControllerBase.ts
+++ b/src/server/plugins/engine/pageControllers/PageControllerBase.ts
@@ -147,7 +147,8 @@ export class PageControllerBase {
           labelOrLegend.isPageHeading = isPageHeading
 
           // Check for optional in label
-          const isOptional = labelOrLegend.text.includes('optional')
+          const isOptional =
+            this.components.formItems.at(0)?.options.required === false
 
           if (pageTitle) {
             labelOrLegend.text = isOptional
--- a/src/server/plugins/engine/pageControllers/PageControllerBase.ts
+++ b/src/server/plugins/engine/pageControllers/PageControllerBase.ts
@@ -146,8 +146,13 @@ export class PageControllerBase {
         if (isPageHeading) {
           labelOrLegend.isPageHeading = isPageHeading
 
+          // Check for optional in label
+          const isOptional = labelOrLegend.text.includes('optional')
+
           if (pageTitle) {
-            labelOrLegend.text = pageTitle
+            labelOrLegend.text = isOptional
+              ? pageTitle + ' (optional)'
+              : pageTitle
           }
 
           pageTitle = pageTitle || labelOrLegend.text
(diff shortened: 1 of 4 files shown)


src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -146,8 +146,13 @@ export class PageControllerBase {
...
           labelOrLegend.isPageHeading = isPageHeading
 
+          // Check for optional in label
+          const isOptional = labelOrLegend.text.includes('optional')
+
           if (pageTitle) {
-            labelOrLegend.text = pageTitle
+            labelOrLegend.text = isOptional
+              ? pageTitle + ' (optional)'
+              : pageTitle
```
KiranVarma02, line 155: I am not convinced this is best solution but I didn't find any other way to do it. 

@alexluckett , @davidjamesstone  - `heading.html` is not the file used to render from my investigation. govukInput in `textfield.html` renders label, so I made above change. 

This solution works. 

Please let me know if I missed anything or better solution.

KiranVarma02: Demo'd and tested few components with Naz. All looks good.

@alexluckett

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -146,8 +146,14 @@ export class PageControllerBase {
         if (isPageHeading) {
           labelOrLegend.isPageHeading = isPageHeading
 
+          // Check for optional in label
+          const isOptional =
+            this.components.formItems.at(0)?.options.required === false
```
colinrotherham, line 151: Is this the same as `formComponents[0]` above?

Sorry not checked properly! Thanks @KiranVarma02

```suggestion
            formComponents[0].options.required === false
```

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -146,8 +146,14 @@ export class PageControllerBase {
...
+
           if (pageTitle) {
-            labelOrLegend.text = pageTitle
+            labelOrLegend.text = isOptional
+              ? pageTitle + ' (optional)'
```
colinrotherham, line 155: Mind using the export from **src/server/plugins/engine/components/constants.ts**?

```suggestion
              ? `${pageTitle}${optionalText}`
```

src/server/plugins/engine/pageControllers/PageControllerBase.ts
KiranVarma02: No, its not same. formComponents doesn't have options data.

src/server/plugins/engine/pageControllers/PageControllerBase.ts
KiranVarma02: done

test/condition/text.test.js
```diff
@@ -33,7 +33,7 @@ describe('TextField based conditions', () => {
     const { container } = await renderResponse(server, options)
 
     const $input = container.getByRole('textbox', {
-      name: 'First page'
+      name: 'First page (optional)'
```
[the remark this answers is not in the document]
KiranVarma02: done

alexluckett (approved):

PR #477, opened 2024-10-18
chore: No restricted imports lint

KiranVarma02:
To save time during code reviews

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -86,13 +86,13 @@ module.exports = {
           }
         ],
 
-        // Check relative import paths are used
+        // Check relative import paths use aliases
         'no-restricted-imports': [
           'error',
           {
             patterns: [
               {
-                group: ['../'],
+                group: ['./', '../'],
                 message: 'Relative imports are not allowed.'
               }
             ]
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -93,7 +93,7 @@ module.exports = {
             patterns: [
               {
                 group: ['./', '../'],
-                message: 'Relative imports are not allowed.'
+                message: "Please use '~/*' import alias instead."
               }
             ]
           }
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -86,6 +86,19 @@ module.exports = {
           }
         ],
 
+        // Check relative import paths are used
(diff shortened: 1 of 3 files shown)


.eslintrc.cjs
```diff
@@ -86,6 +86,19 @@ module.exports = {
           }
         ],
 
+        // Check relative import paths are used
+        'no-restricted-imports': [
+          'error',
+          {
+            patterns: [
+              {
+                group: ['../'],
```
colinrotherham, line 95: Nice one. What about sibling imports?

```suggestion
                group: ['./', '../'],
```

KiranVarma02: I thought we use `./` in some places? let me check

.eslintrc.cjs
```diff
@@ -86,6 +86,19 @@ module.exports = {
           }
         ],
 
+        // Check relative import paths are used
```
colinrotherham, line 89: I think we're making sure we _don't_ use relative imports?

We're trying to stick with `~` aliases instead (so files can be moved without changing their imports)

```suggestion
        // Check relative import paths use aliases
```

.eslintrc.cjs
KiranVarma02: in 1 place :), I will correct that and add it to the rule

KiranVarma02: done

.eslintrc.cjs
```diff
@@ -86,6 +86,19 @@ module.exports = {
...
         ],
 
+        // Check relative import paths use aliases
+        'no-restricted-imports': [
+          'error',
+          {
+            patterns: [
+              {
+                group: ['./', '../'],
+                message: 'Relative imports are not allowed.'
```
colinrotherham, line 96: Sorry, I forgot this message also said relative imports not aliases

Unless the default is obvious enough and we don't need one?

Thanks @KiranVarma02

```suggestion
                message: "Please use '~/*' import alias instead.'
```

KiranVarma02: perfect

colinrotherham (approved):
Thanks @KiranVarma02

Want to keep this open and open PRs across the other repos too?

Can come back to it if we discover any other tweaks are needed

PR #484, opened 2024-10-18
summary changes complete

KiranVarma02:


src/server/views/summary.html
```diff
@@ -18,6 +19,13 @@
         }) }}
       {% endif %}
 
+      {% if notificationEmail is undefined %}
```
colinrotherham, line 22: Probably good to check for empty values too

```suggestion
      {% if not notificationEmail | length %}
```

KiranVarma02: agree, done

src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -119,13 +119,22 @@ export class SummaryPageController extends PageController {
         }
       }
 
+      const { params } = request
+      const { slug } = params as { slug: string }
```
colinrotherham, line 123: Mind if we hold off using `as` for now?

You'll have `params.slug` working as expected after https://github.com/DEFRA/forms-runner/pull/469/commits/4945bb5043312d3b2ee83af25a6ce8465eddef85 is merged

src/server/plugins/engine/views/common/warn-missing-notification-email.html
```diff
@@ -0,0 +1,11 @@
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+
+{% macro warnMissingNotificationEmail(notificationEmail, designerUrl, slug) %}
+  {% if not notificationEmail | length %} 
+    {{ govukWarningText({
+        html: 'If you want to receive a test email, <a target="_blank" class="govuk-link" href=' + designerUrl + '/library/' + slug + '/edit/notification-email>enter the email address (opens in new tab)</a> you want form submissions to go to.',
+        iconFallbackText: "Warning"
+    }) }}
+  {% endif %}
+{% endmacro %}
```
KiranVarma02, line 10: changing macro to include is also ok but I think I prefer this. Let me know what you think!

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -472,6 +476,18 @@ export class PageControllerBase {
 
       viewModel.backLink = this.getBackLink(progress)
 
+      if (progress[0]) {
```
KiranVarma02, line 479: This is not correct. How can I correctly identify first page to satisfy below

`Should appear on the first question page when previewing a form and the check your answers page if email address is missing.` 

https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/419250#19509961

KiranVarma02: I was testing wrong scenario. This issue is resolved.

src/server/plugins/engine/views/common/warn-missing-notification-email.html
colinrotherham: Nice one @KiranVarma02

We'll need to restrict this to "preview mode" otherwise there's a possibility our link to **forms-designer** will appear to users in production on live-but-misconfigured forms

For your macro versus partial question, shall we keep it as a partial like the others?

We've got `slug` etc available via the default context so they don't need to be macro params:

https://github.com/DEFRA/forms-runner/blob/9ba19cb625ebfc7b56b220c6a722f9667a7b5880/src/server/plugins/views.ts#L66-L72

src/server/plugins/engine/views/common/warn-missing-notification-email.html
```diff
@@ -0,0 +1,8 @@
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+
+{% macro warnMissingNotificationEmail(designerUrl, slug) %}
+    {{ govukWarningText({
+        html: 'If you want to receive a test email, <a target="_blank" class="govuk-link" href=' + designerUrl + '/library/' + slug + '/edit/notification-email>enter the email address (opens in new tab)</a> you want form submissions to go to.',
```
colinrotherham, line 5: Can you use a separate `{% set html %}` for the `html` property like in [**preview-banner.html**](https://github.com/DEFRA/forms-runner/blob/main/src/server/plugins/engine/views/partials/preview-banner.html)?

Makes it a bit easier/safer than string joining HTML

src/server/plugins/engine/views/index.html
```diff
@@ -14,6 +17,10 @@
         {{ govukErrorSummary(errors) }}
       {% endif %}
 
+      {% if isStartPage %}
+        {{ warnMissingNotificationEmail(designerUrl, slug) }}
+      {% endif %}
```
colinrotherham, line 22: What if someone clicks "Preview" (in **forms-designer**) on a question page?

They'll miss the warning. This might be a requirement, just making sure

src/server/plugins/engine/views/common/warn-missing-notification-email.html
colinrotherham: You might like what we do in **forms-designer**

We set up a cut-down [`config` context property](https://github.com/DEFRA/forms-designer/blob/c7dd495cd4f9eb8cefd95d175648527100cb6d97/designer/server/src/common/nunjucks/context/index.js#L35-L38) which gives `{{ config.serviceName }}` in Nunjucks

src/server/plugins/engine/views/index.html
alexluckett: Just confirming this is the desired behavior from the story. First and summary page only.

src/server/plugins/engine/views/index.html
```diff
@@ -14,6 +17,10 @@
         {{ govukErrorSummary(errors) }}
       {% endif %}
 
+      {% if isStartPage %}
```
alexluckett, line 20: You're already doing a check in the page controller, but the check is repeated here. What if the two drift?

Earlier, you check if it's a start page or a summary page. But here you're only checking if it's a start page. 

I'd keep the check only in the page controller as you did earlier, but just send a flag rather than the individual checks. e.g. `{% if isStartPage %}` would become `{% if isMissingNotificationEmail %}`. isMissingNotificationEmail would be computed in the controller based on `isStartPage || isSummaryPage`.

src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -154,6 +154,13 @@ export class SummaryPageController extends PageController {
 
       viewModel.backLink = this.getBackLink(progress)
 
+      const missingEmailWarning =
+        await this.buildMissingEmailWarningModel(request)
+
+      if (missingEmailWarning) {
+        return h.view('summary', { ...viewModel, ...missingEmailWarning })
```
alexluckett, line 161: we already have a `viewModel`. Why not include these attributes on that `viewModel` rather than creating a new view model containing other view models?

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -463,10 +466,36 @@ export class PageControllerBase {
...
       return h.view(this.viewName, viewModel)
     }
   }
 
+  async buildMissingEmailWarningModel(request: Request, isStartPage?: boolean) {
+    const { params } = request
+    // Warn the user if the form has no notification email set only on start page and summary page
+    if (isStartPage || params.path === 'summary') {
+      const { slug } = params as { slug: string }
+      const { notificationEmail } = await getFormMetadata(slug)
```
alexluckett, line 487: Glad to see we're not wasting API calls on every page request and only doing it when required

src/server/plugins/engine/pageControllers/SummaryPageController.ts
KiranVarma02: I thought about it and the reason I didn't do it is - it didn't feel right to add form metadata to actual form model. Let me know what you think!

src/server/plugins/engine/views/index.html
KiranVarma02: the reason I have check in page controller is to not waste API calls like you said in other comment.

Yes, that's correct because we have different view for summary and for everything else and same model is used for both summary and other pages. 

I think its more efficient the way its done now. I only need to check for notificationEmail for first page. I don't need to do `{% if isMissingNotificationEmail %}` check for every page. Just for more clarify, I can add `notification email` check to `isStartPage` check but it wouldn't make any difference to functionality. 

Happy to get on a call if needed

src/server/plugins/engine/pageControllers/PageControllerBase.ts
colinrotherham, line 487: Just linking to https://github.com/DEFRA/forms-runner/pull/484#discussion_r1808802631 because the code has moved

```suggestion
      const { notificationEmail } = await getFormMetadata(params.slug)
```

src/server/plugins/engine/views/common/warn-missing-notification-email.html
KiranVarma02: done

src/server/plugins/engine/pageControllers/PageControllerBase.ts
KiranVarma02: I am also using slug in line 491. I need it in a const.

KiranVarma02: done

src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -158,12 +158,10 @@ export class SummaryPageController extends PageController {
...
       viewModel.backLink = this.getBackLink(progress)
 
-      const missingEmailWarning =
+      const notificationEmailWarning =
         await this.buildMissingEmailWarningModel(request)
 
-      if (missingEmailWarning) {
-        return h.view('summary', { ...viewModel, ...missingEmailWarning })
-      }
+      viewModel.notificationEmailWarning = notificationEmailWarning
```
alexluckett: `notificationEmailWarning` isn't even neccessary as a variable any more, you could directly assign the result to `viewModel.notificationEmailWarning`.

alexluckett (commented):
You just need some tests to cover this scenario and you're good to go.

src/server/plugins/engine/views/index.html
alexluckett: Discussed over a call. Resolved.

src/server/plugins/engine/pageControllers/SummaryPageController.ts
alexluckett: DIscussed over a call. Resolved.

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -473,10 +476,36 @@ export class PageControllerBase {
 
       viewModel.backLink = this.getBackLink(progress)
 
+      const notificationEmailWarning = await this.buildMissingEmailWarningModel(
+        request,
+        isStartPage
+      )
+
+      viewModel.notificationEmailWarning = notificationEmailWarning
```
alexluckett, line 484: ditto for my earlier comment, you've already done this for the summary page controller 🙂

KiranVarma02: I thought to do it but didn't do it not to put all that in 1 line but ok :)

alexluckett (commented):
I just did some manual testing. It throws an error when a preview is submitted without an output email. It shouldn't send the email if there's no email defined.

`/src/server/plugins/engine/pageControllers/SummaryPageController.ts:201` will need amending + a test.

If you have an email, send the email. If there's no email and it's a live form, throw the existing error.

src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -231,16 +232,19 @@ async function submitForm(
...
+  if (emailAddress) {
+    await sendEmail(request, summaryViewModel, model, emailAddress)
+  }
 }
 
 async function extendFileRetention(
   model: FormModel,
   state: FormSubmissionState,
-  updatedRetrievalKey: string
+  updatedRetrievalKey: string | undefined
```
KiranVarma02, line 247: Can i do something without having to accept undefined?

alexluckett (commented):
The unit tests are good, but we don't have a journey test covering this whole scenario (the code is dotted all around the codebase). I'd recommend a new `test/form/submission-email.test.js` that sets up a form with/without the output email and validates how it behaves on start/summary.

KiranVarma02:
> The unit tests are good, but we don't have a journey test covering this whole scenario (the code is dotted all around the codebase so a single unit test doesn't cover it). I'd recommend a new `test/form/submission-email.test.js` that sets up a form with/without the output email and validates how it behaves on start/summary to prove the requirement is met.

Summary page - with email/without email:
https://github.com/DEFRA/forms-runner/pull/484/files#diff-752d39d82bff37b0e548917c3f3da52af08442f7d7af1aee4ab92746100ff678R64-R100

First page - without email
https://github.com/DEFRA/forms-runner/pull/484/files#diff-a94a1e217aec1b54d86e919033a12dc09182e0b61832d1c4b45f718ff91cd15cR47-R51

I will add first page with email. 

Do you think we need a new test file or this is good?

src/server/plugins/engine/services/formSubmissionService.js
```diff
@@ -6,7 +6,7 @@ const submissionUrl = config.get('submissionUrl')
 /**
  * Persist files by extending the time-to-live to 30 days
  * @param {{fileId: string, initiatedRetrievalKey: string}[]} files - batch of files to persist
- * @param {string} persistedRetrievalKey - final retrieval key when submitting
+ * @param {string | undefined} persistedRetrievalKey - final retrieval key when submitting
```
alexluckett, line 9: this can go back to just `string` as you're not calling it with undefined any more.

alexluckett (approved):

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -463,10 +466,36 @@ export class PageControllerBase {
...
+
+      if (missingEmailWarning) {
+        return h.view(this.viewName, { ...viewModel, ...missingEmailWarning })
+      }
+
       return h.view(this.viewName, viewModel)
     }
   }
 
+  async buildMissingEmailWarningModel(request: Request, isStartPage?: boolean) {
```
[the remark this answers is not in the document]
KiranVarma02: My understanding is requirement is only when we load the page. If this is a data issue then yes I agree it has to be in POST check but I am not sure in this case. 

@alexluckett

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -463,10 +466,36 @@ export class PageControllerBase {
 
       viewModel.backLink = this.getBackLink(progress)
 
+      const missingEmailWarning = await this.buildMissingEmailWarningModel(
```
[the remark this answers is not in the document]
KiranVarma02: I didn't see this, I agree. 

I will do it as a different PR.

PR #487, opened 2024-10-22
Add missing components to GOV.UK Notify tests

colinrotherham:
Closes assurance ticket [#461367](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/461367)

Adds tests for the following components:

* Autocomplete
* Month & Year

Plus tests for the "Form received at" date and "Download all" link

### DIFF
--- a/test/form/govuk-notify.test.js
+++ b/test/form/govuk-notify.test.js
@@ -140,6 +140,8 @@ Arabian,Shire,Race
 
 * [test.pdf](https://test-designer.cdp-int.defra.cloud/file-download/5a76a1a3-bc8a-4bc0-859a-116d775c7f15)
 
+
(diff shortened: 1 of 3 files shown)


test/form/govuk-notify.test.js
```diff
@@ -94,6 +105,12 @@ privateLimitedCompany
 \`\`\`
 
 
+## Autocomplete field
+\`\`\`
+910400044
+\`\`\`
```
colinrotherham, line 111: Unlike checkboxes and radios etc, the raw value `910400044` is output not "Czech Republic"

Should we change this?

alexluckett: I would prefer the human readable label personally. I've flagged it with Naz.

alexluckett (approved):

PR #504, opened 2024-10-29
Fix clash on `FormState` type name

colinrotherham:
Quick PR to rename the new `FormState` enum to `FormStatus` and use it across the repo

It clashed with the `FormState` type we use for Redis session state

Fixes [bug #465015](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/465015)


KiranVarma02 (commented):
LGTM. If you can satisfy sonar that will be great.

KiranVarma02 (approved):

PR #507, opened 2024-10-29
Add environment identification tag in the header (409424)

KiranVarma02:
Adds an environment tag to the header

Production tag added for reference but will be hidden on **forms-runner**

<br>

Header tag development

Header tag performance test

Header tag production


src/config/index.ts
```diff
@@ -27,6 +27,11 @@ export const config = convict({
     env: 'PORT'
   },
   env: {
+    format: ['development', 'test', 'pref test', 'production'],
```
colinrotherham, line 30: Watch out for the `pref test` typo

Think we should match them to the actual env names? [See comment](https://github.com/DEFRA/forms-designer/pull/483/files#r1825666648)

```suggestion
    format: ['dev', 'test', 'perf-test', 'prod'],
```

src/server/views/layout.html
```diff
@@ -45,7 +44,7 @@
       </a>
     
     {% if envTag %}
```
colinrotherham, line 46: Do we need `envTag` everywhere?

```suggestion
    {% if config.env !== "prod" %}
```

src/server/views/tag-env/template.njk
```diff
@@ -0,0 +1,15 @@
...
+    {% case "production" %}
+      {% set classes = "govuk-tag--red" %}
+    {% default %}
+      {% set classes = "govuk-tag--grey" %}
+  {% endswitch -%}
+
+  {{ govukTag({
+    text: env | capitalize,
+    classes: classes
+  }) }}
```
colinrotherham, line 15:
```suggestion
  {% case "test" %}
    {% set classes = "govuk-tag--yellow" %}
  {% case "production" %}
    {% set classes = "govuk-tag--red" %}
  {% default %}
    {% set classes = "govuk-tag--grey" %}
{% endswitch -%}

{{ govukTag({
  text: env | capitalize,
  classes: classes
}) }}
```

src/config/index.ts
KiranVarma02: corrected typo. I will wait for @alexluckett to confirm regarding env names, same for designer PR

src/server/views/tag-env/template.njk
colinrotherham: Same 😭

KiranVarma02: same? are you sure? did you pull latest?

colinrotherham: Ahh sorry, no I mean I also wish we had a linter for it! 😅

KiranVarma02:
is there something we can do about sonar major issue? Its same in designer. I guess this can be ignored?

alexluckett (approved):

PR #508, opened 2024-10-31
Fix number `schema.precision` not being applied etc

colinrotherham:
This PR closes [bug #414012](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/414012) and includes:

* Number field `schema.precision` is applied
* Number field error messages are consistent
* Date field decimal places are removed

## Number fields
When empty, we said "XXX must be a number" rather than the default "Enter XXX"

### Before

Number field before

### After

Number field after

## Date fields
For all child number fields used by date or month/year field, we didn't include the parent label in the message

### Before

Date input before

### After

Date input after


KiranVarma02:
Looks ok to me once all checks are passing

alexluckett (approved):

PR #509, opened 2024-11-01
458970 Validate nested fields when optional but partially entered

colinrotherham:
This PR fixes an existing bug as part of [story #458970](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/458970)

E.g. If you set a Month/Year or UK Address field to “optional” you can store partial values in the state:

* Month but no Year
* Year but no Month

This is because the `ComponentCollection` children are all valid—they’re checked one-by-one not as a whole

This fix is necessary as session changes in PR https://github.com/DEFRA/forms-runner/pull/470 start affecting the Date field too

## Joi object peers

This is fixed by adding `peers` to a Joi schema

>[`object.and(...peers, [options])`](https://joi.dev/api/?v=17.13.3#objectandpeers-options) — Defines an all-or-nothing relationship between keys where if one of the peers is present, all of them are required as well

```js
formSchema = formSchema.and(
  `${name}__day`,
  `${name}__month`,
  `${name}__year`,
  {
    isPresent: isFormValue
  }
)
```

### DIFF
--- a/src/server/plugins/engine/components/ComponentCollection.ts
+++ b/src/server/plugins/engine/components/ComponentCollection.ts
@@ -1,7 +1,6 @@
 import { type ComponentDef } from '@defra/forms-model'
-import joi, { type ObjectSchema } from 'joi'
+import joi, { type CustomValidator, type ObjectSchema } from 'joi'
 
-import { type ComponentSchemaNested } from '~/src/server/plugins/engine/components/ComponentBase.js'
 import {
   createComponentField,
   type ComponentFieldClass,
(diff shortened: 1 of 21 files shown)


alexluckett (approved):

src/server/plugins/engine/components/ComponentCollection.ts
```diff
@@ -135,4 +196,39 @@ export class ComponentCollection {
...
+   * Return error by code for collection fields
+   * For example, 'date.min', 'date.max'
+   */
+  error(helpers: CustomHelpers, code: string, context?: Context) {
+    const { formItems, parent } = this
+    const [field] = formItems
+
+    // Use collection parent name/title when available
+    const key = parent?.name ?? field.name
+    const title = parent?.title ?? field.title
```
colinrotherham, line 210: But similarly to https://github.com/DEFRA/forms-runner/pull/509/files#r1827483136 we also need to locate the parent component so we can pass `{{#title}}` into custom error messages should it be needed:

```js
'{{#title}} must include a {{#label}}'
```

src/server/plugins/engine/components/ComponentCollection.ts
```diff
@@ -135,4 +196,39 @@ export class ComponentCollection {
...
+    // Support collection parent {{#title}} in messages
+    const local: Context = {
+      title: title.toLowerCase(),
+      ...context
+    }
+
+    // Error summary links to first collection field
+    const localState: State = parent
+      ? { key, path: [parent.name, field.name] } // For example, `#dateField__day`
+      : { key, path: [field.name] } // Otherwise `#dateField` for single fields
```
colinrotherham, line 221: I've also fixed this issue where `href="#dateField"` in the error summary wouldn't link anywhere

Using the parent field, we now locate the first `href="#dateField__day"` when the entire date is invalid

E.g. When validating impossible dates such as 31/2/2024

davidjamesstone (approved):

PR #517, opened 2024-11-04
Remove optional text from fields when shown in legend

colinrotherham:
This PR closes [bug #468636](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/468636)

It partially reverts a previous change to add "**(optional)**" to form fields with legends

## Before
The optional text is shown in the legend and child fields

Optional before

## After
The optional text is shown only in the legend unless the component title is hidden

Optional after

### DIFF
--- a/src/server/plugins/engine/components/DatePartsField.ts
+++ b/src/server/plugins/engine/components/DatePartsField.ts
@@ -8,7 +8,6 @@ import {
   isFormState
 } from '~/src/server/plugins/engine/components/FormComponent.js'
 import { NumberField } from '~/src/server/plugins/engine/components/NumberField.js'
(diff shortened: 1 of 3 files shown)


src/server/plugins/engine/components/DatePartsField.ts
```diff
@@ -47,7 +45,7 @@ export class DatePartsField extends FormComponent {
           schema: { min: 1, max: 31, precision: 0 },
           options: {
             required: isRequired,
-            optionalText: !isRequired && hideOptional,
+            optionalText: true,
```
colinrotherham, line 48: I've hard coded `optionalText` here rather than a previous workaround to `.replace()` it in the view model

davidjamesstone (approved):

PR #515, opened 2024-11-04
Share props between single and collection fields

colinrotherham:
This is a tidy up PR to align collection fields with component fields to avoid confusion

All fields now have `this.parent` to determine if they're part of a collection

## Single fields

* Added `this.parent` to match collection fields
* Renamed view model `children` to `components`
* Updated constructor to `new FormComponent(def, { parent, model }})`

## Collection fields

* Renamed `items` to `components`
* Renamed `formitems` to `fields`
* Added `this.parent` in [#509](https://github.com/DEFRA/forms-runner/pull/509)
* Updated constructor in [#509](https://github.com/DEFRA/forms-runner/pull/509) to `new ComponentCollection(def, { parent, model }, schema})`

PR #518, opened 2024-11-04
Add number precision error messages

colinrotherham:
This PR closes [bug #414012](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/414012) and adds error messages for:

* `number.precision`
* `number.integer`
* `number.unsafe`

## Whole numbers
When number fields are set to `precision: 0` we validate whole numbers only

Number integer error message

## Decimal numbers
When number fields are set to `precision: 2` etc we validate maximum decimal places

Number precision error message


src/server/plugins/engine/components/NumberField.ts
```diff
@@ -23,7 +23,11 @@ export class NumberField extends FormComponent {
 
     const { options, schema, title } = def
 
-    let formSchema = joi.number().label(title.toLowerCase()).required()
+    let formSchema = joi
+      .number()
+      .custom(getValidatorPrecision(this))
```
colinrotherham, line 28: Preventing rounding using `.prefs({ convert: false })` made every "string number" invalid 😔

I've used a custom validator instead so we can keep string-to-number conversion

src/server/plugins/engine/components/NumberField.ts
```diff
@@ -111,3 +117,31 @@ export class NumberField extends FormComponent {
     return typeof value === 'number'
   }
 }
+
+export function getValidatorPrecision(component: NumberField) {
```
mokhld, line 121: Prob being pedantic here (sorry!), but this func is returning a validator so maybe `getPrecisionValidator` is more apt? Initially reading this I thought the function returns a precision value.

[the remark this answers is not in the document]
mokhld: Sorry ignore me, saw that you've typed it already below 😅

mokhld (approved):

src/server/plugins/engine/components/NumberField.ts
colinrotherham: Yeah it makes sense now 😆

We have `getValidatorXXX` elsewhere but if you fancy changing things like this in another PR?

E.g. I'm slowly tackling confusing names in this one for example:

* https://github.com/DEFRA/forms-runner/pull/515

PR #530, opened 2024-11-11
Use standard text type for NumberField

davidjamesstone:


src/server/plugins/engine/components/NumberField.ts
```diff
@@ -82,8 +82,13 @@ export class NumberField extends FormComponent {
     const viewModel = super.getViewModel(payload, errors)
     let { attributes, prefix, suffix, value } = viewModel
 
-    if (schema.precision) {
-      attributes.step = '0.' + '1'.padStart(schema.precision, '0')
+    if (
+      !schema.precision ||
+      (NumberField.isNumber(schema.precision) && schema.precision <= 0)
+    ) {
```
colinrotherham, line 88: Can use `this.isValue()` versus `NumberField.isNumber()` when inside the class

But might be better reserving the **isValue** and **isState** helpers for payload/state only?

```suggestion
    if (typeof schema.precision === 'undefined' || schema.precision <= 0) {
```

src/server/plugins/engine/components/NumberField.test.ts
```diff
@@ -190,7 +189,37 @@ describe('NumberField', () => {
...
+        const viewModel = componentCustom.getViewModel(getFormData(99))
+
+        expect(viewModel.attributes).toEqual(
+          expect.objectContaining({
+            inputmode: 'numeric'
+          })
+        )
+      })
+
+      it("Doesn't set Nunjucks component inputmode attribute when precision is positive", () => {
```
colinrotherham, line 222: Sorry, should probably find a Jest ESLint rule that keeps these lowercase

Cheers for all these tests

```suggestion
      it("does not set Nunjucks component inputmode attribute when precision is positive", () => {
```

src/server/plugins/engine/components/NumberField.ts
davidjamesstone: Resolved with 25530c8557fc1d29375036effe886994e9692944

colinrotherham: Makes sense if we hit this more in future, cheers

colinrotherham:
Interesting research on why we can't use `inputmode="decimal"`

* https://github.com/alphagov/govuk-design-system/pull/1279#issuecomment-639467489

src/server/plugins/engine/components/NumberField.test.ts
```diff
@@ -199,8 +228,8 @@ describe('NumberField', () => {
         const viewModel = componentCustom.getViewModel(getFormData(99.99))
 
         expect(viewModel.attributes).toEqual(
-          expect.objectContaining({
-            step: '0.01'
+          expect.not.objectContaining({
+            inputmode: 'numeric'
           })
         )
```
colinrotherham, line 234: They're definitely a lot shorter

```suggestion
        expect(viewModel.attributes).not.toHaveProperty('inputmode')
```

colinrotherham (approved):

src/server/plugins/engine/components/NumberField.test.ts
```diff
@@ -190,7 +189,37 @@ describe('NumberField', () => {
...
+          formModel
+        )
+
+        const viewModel = componentCustom.getViewModel(getFormData(99))
+
+        expect(viewModel.attributes).toEqual(
+          expect.objectContaining({
+            inputmode: 'numeric'
+          })
+        )
```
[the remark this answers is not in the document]
davidjamesstone: Resolved with 9a7bb29af2bf397d9d626add77ecb1fe3d0ff25d

PR #542, opened 2024-11-18
bug/475453 add another backlink

davidjamesstone:


mokhld (approved):

colinrotherham (approved):
Looks brilliant

Any chance we can add a "Back link" selector to check the href?

It's how the bug was reported to us, good to close that gap

```js
const $backLink = container.getByRole('link', {
  name: 'Back'
})

expect($backLink).toBeInTheDocument()
expect($backLink).toHaveAttribute('href', XXXX)
```

alexluckett (approved):

src/server/plugins/engine/pageControllers/RepeatPageController.ts
```diff
@@ -179,6 +177,12 @@ export class RepeatPageController extends PageController {
       request: FormRequest,
       h: ResponseToolkit<FormRequestRefs>
     ) => {
+      if (!request.params.itemId) {
+        return h
+          .redirect(`${request.path}/${randomUUID()}${request.url.search}`)
+          .code(StatusCodes.SEE_OTHER)
+      }
+
```
colinrotherham, line 185: Thanks for adding the tests @davidjamesstone, coverage has jumped up now 🙌

Have you noticed we don't have coverage when `getRepeatAppData()` throws?

Wonder if that would be a good place to move this redirect? Instead of throwing

```js
throw badRequest('No repeat data found on the request')
```

Merge away, not a blocker

davidjamesstone: `throw badRequest('No repeat data found on the request')` is only there for typescript narrowing purposes. `request.app.repeat` should never be undefined in `getAppRepeatData` (assuming `setAppRepeatData` has been called prior).

To me the top of `makeGetRouteHandler` is the natural place for the redirect to occur and while `request.app.repeat` should never be undefined, if it ever is and we change to do the redirect in that block, there's an increased risk of a redirect loop.

PR #545, opened 2024-11-19
Add test to ensure FormModel doesn't throw when conditions are passed with apostrophes

davidjamesstone:


colinrotherham (approved):
This is good to go btw once the conflict is sorted

Don't have to block on [#546](https://github.com/DEFRA/forms-runner/pull/546) as I can address it there instead

PR #546, opened 2024-11-20
Add support for ES module form definitions

colinrotherham:
This PR extends **forms-runner** to support `*.js` form definitions alongside `*.json`

I've split out [#477926](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/477926) as part of [story #458979](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/458979) because it's quite fragile to keep copying & pasting JSON files

```js
const server = await createServer({
  formFileName: 'example.js',
  formFilePath: join(cwd(), 'server/forms'),
})
```

E.g. We can now import a shared `file-upload.js` definition but tweak it for each assertion

Without losing type safety


src/server/plugins/engine/configureEnginePlugin.ts
```diff
@@ -1,38 +1,45 @@
...
 }
+
+export async function getForm(importPath: string) {
+  const { ext } = parse(importPath)
+
+  const attributes: ImportAttributes = {
+    type: ext === '.json' ? 'json' : 'module'
+  }
+
+  const formImport = import(importPath, { with: attributes }) as Promise<{
```
mokhld, line 39: @colinrotherham do we need to gracefully handle any errors here if the import fails?

colinrotherham: Nah all good, we don't currently (see below) and we only load JSON files in tests

In production we load forms from **forms-manager** via API

https://github.com/DEFRA/forms-runner/blob/fadd272e674a2f369645dd01229b81d9ecee15c8/src/server/plugins/engine/configureEnginePlugin.ts#L23-L25

mokhld: ah okay, that makes sense.

mokhld (approved):
Really great job on this @colinrotherham 🎆 !

PR #550, opened 2024-11-20
Add test coverage for conditional next page lookup

colinrotherham:
This PR adds code coverage for conditional page lookup

I've also moved `controller.getSummaryPath()` to the base class

Identified as part of [story #458979](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/458979) but split into [#477910](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/477910)


src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -171,29 +171,17 @@ export class PageControllerBase {
...
-        return {
-          ...next,
-          page
-        }
-      })
-      .filter((v) => !!v)
+    // Remove stale links
+    return pageDef.next.filter(({ path }) =>
+      def.pages.some((page) => path === page.path)
+    )
```
colinrotherham, line 184: Rather than loop instantiated `model.pages` we can use `def.pages` instead (lighter)

davidjamesstone (approved):

PR #557, opened 2024-11-26
Fix the "Jenny bug"

colinrotherham:
This PR fixes the "Jenny bug" and removes the "Enter XXX" missing field workaround

We add both `item.href` and `item.error` on **detail items** so can easily redirect back to pages with errors:

```js
if (errorField) {
  return h.redirect(errorField.href)
}
```

**Note:** Invalid state from missed pages is only checked when reaching the `/summary` page

Since exit pages are reached with valid state the summary page can still be accessed and submitted

---

This is part of a series of PRs where 2) and 3) below both fix and improve the ["exit page bug" #482470](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/482470)

1. **“Jenny bug” fix** in [#557](https://github.com/DEFRA/forms-runner/pull/557) (this PR)
Validating state at the end and redirecting to pages with errors

2. **Fast walking with direct page access** in [#568](https://github.com/DEFRA/forms-runner/pull/568)
Given a landing path, check it's in the list of relevant paths

3. **Slow walking with redirects to correct page** in [#570](https://github.com/DEFRA/forms-runner/pull/570)
Same as 2) but validate state is correct before running conditions on it

### DIFF
--- a/src/server/plugins/engine/pageControllers/SummaryPageController.ts
+++ b/src/server/plugins/engine/pageControllers/SummaryPageController.ts
@@ -126,14 +126,6 @@ export class SummaryPageController extends PageController {
       const state = await this.getState(request)
       const summaryViewModel = this.getSummaryViewModel(state, request)
 
-      // Display error summary on the summary
-      // page if there are incomplete form errors
-      if (summaryViewModel.errors) {
-        summaryViewModel.showErrorSummary = true
-
-        return h.view('summary', summaryViewModel)
-      }
-
       const { params } = request
 
       // Get the form metadata using the `slug` param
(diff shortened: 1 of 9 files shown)


src/server/plugins/engine/pageControllers/SummaryPageController.ts
```diff
@@ -96,45 +95,15 @@ export class SummaryPageController extends PageController {
...
-          return redirectTo(
-            h,
-            `/${model.basePath}${pageWithError.path}`,
-            params
-          )
+        const errorField = viewModel.details
+          .flatMap(({ items }) => items.find(({ error }) => error) ?? [])
+          .at(0)
+
+        if (errorField) {
```
mokhld, line 105: @colinrotherham If it defaults to an empty array (which is truthy) this block will fire. Obv you know this lol, just confirming that's the intended behaviour?

colinrotherham: Ah I panicked then, there's a little `.at(0)` at the end to grab the first item

alexluckett (commented):
Clever solution. I'd like to take a look at this in a bit more detail before merging though, please hold off.

alexluckett (approved):

PR #567, opened 2024-12-03
Prefer page path methods for start/summary

colinrotherham:
This PR adds new page controller getters/methods to assist when walking the form journey

* `page.href`
* `page.getStartPath()`

Mainly to avoid this everywhere:

```patch
- return `/${this.model.basePath || ''}${page.path}`
+ return page.href
```

I've also simplified some page methods where possible

## Test tweaks

I've also added `basePath` to form tests and removed the consts we added to appease Sonar:

```patch
- const okStatusCode = 200
- const redirectStatusCode = 302
- const badRequestStatusCode = 400
```

### DIFF
--- a/src/server/index.test.ts
+++ b/src/server/index.test.ts
@@ -1,4 +1,5 @@
 import { type Server } from '@hapi/hapi'
+import { StatusCodes } from 'http-status-codes'
 
 import { createServer } from '~/src/server/index.js'
 import {
@@ -51,7 +52,7 @@ describe('Model cache', () => {
 
       const res = await server.inject(options)
 
-      expect(res.statusCode).toBe(302)
+      expect(res.statusCode).toBe(StatusCodes.MOVED_TEMPORARILY)
       expect(res.headers.location).toBe('/slug/page-one')
       expect(getCacheSize()).toBe(1)
     })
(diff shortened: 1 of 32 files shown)


src/server/index.test.ts
```diff
@@ -10,10 +10,6 @@ import * as fixtures from '~/test/fixtures/index.js'
 
 jest.mock('~/src/server/plugins/engine/services/formsService.js')
 
-const okStatusCode = 200
-const redirectStatusCode = 302
-const notFoundStatusCode = 404
```
mokhld, line 15: @colinrotherham out of interest, why did you remove this? Seems to make sense to keep it DRY?

mokhld: Sorry ignore my question. I saw your comment about Sonar being the culprit.

colinrotherham: Yeah agreed. We should swap them all for `StatusCodes` at some point

src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -169,6 +167,11 @@ export class PageControllerBase {
     }
   }
 
+  get href() {
+    const { model, path } = this
+    return `/${model.basePath}${path}`
```
mokhld, line 172: Is there ever a case where we might have malformed URLs?

mokhld (approved):

src/server/plugins/engine/pageControllers/PageControllerBase.ts
colinrotherham: Possibly as we support form definitions with `startPage` starting `http(s)://`

Not due to this PR, but we may have potential redirect issues lurking somewhere

mokhld (approved):

PR #571, opened 2024-12-06
Add Google Analytics

alexluckett:


src/server/plugins/router.ts
```diff
@@ -60,6 +60,34 @@ export default {
...
+
+          if (decision === 'yes') {
+            request.yar.set('cookieConsent', true)
+          } else if (decision === 'no') {
+            request.yar.set('cookieConsent', false)
+          } else {
+            throw Boom.badRequest('Unknown cookie preference')
+          }
+
+          request.yar.flash('cookieConsentUpdated', true, true)
```
colinrotherham, line 77: Did you see the [`cookies[hide]` example](https://design-system.service.gov.uk/components/cookie-banner/server-side-confirmation/) after updating consent?

Rather than use Yar we can just save each preference name/value in the same cookie

It also means when JavasScript is available we can write to these simple values client-side, preventing us having to POST the preferences form and navigate away from the page

E.g. Given this default cookie preferences object:

```js
const cookies = {
  essential: true,
  analytics: false,
  hide: false
}
```

## Accept analytics cookies
Clicking the accept button:

```html
<button name="cookie[analytics]" value="true">
  Accept analytics cookies
</button>
```

Updates the preferences:

```js
const cookies = {
  essential: true,
  analytics: true,
  hide: false
}
```

## Hide cookie message
Clicking the hide button:

```html
<button name="cookie[hide]" value="true">
  Hide cookie message
</button>
```

Updates the preferences:

```js
const cookies = {
  essential: true,
  analytics: true,
  hide: true
}
```

src/server/views/layout.html
```diff
@@ -43,6 +55,51 @@
...
 {% block header %}
+  {% if googleAnalyticsTrackingId and (cookieConsent !== true and cookieConsent !== false) %}
+    <form method="post" action="/help/cookie-preferences">
+      {% set html %}
+        <p class="govuk-body">We use some essential cookies to make this service work.</p>
+        <p class="govuk-body">We’d like to set additional cookies so we can remember your settings, understand how people use the service and make improvements.</p>
+      {% endset %}
+
+      {% set acceptHtml %}
+        <p class="govuk-body">You’ve accepted additional cookies. You can <a class="govuk-link" href="#">change your cookie settings</a> at any time.</p>
```
colinrotherham, line 66: Watch out for the stray `href="#"` here

src/server/views/layout.html
```diff
@@ -43,6 +55,51 @@
...
+      {% endset %}
+
+      {% set rejectedHtml %}
+        <p class="govuk-body">You’ve rejected additional cookies. You can <a class="govuk-link" href="#">change your cookie settings</a> at any time.</p>
```
colinrotherham, line 70: Another one

src/server/views/layout.html
```diff
@@ -29,6 +31,16 @@
   <link rel="preload" as="font" href="{{ assetPath }}/fonts/bold-b542beb274-v2.woff2" type="font/woff2" crossorigin="anonymous">
   <link rel="preload" as="font" href="{{ assetPath }}/fonts/light-94a07e06a1-v2.woff2" type="font/woff2" crossorigin="anonymous">
   <link rel="stylesheet" href="{{ getAssetPath("stylesheets/application.scss") }}">
+
+  {% if googleAnalyticsTrackingId and cookieConsent === true %}
+    <script data-testid="ga-tag-js-main" async src="https://www.googletagmanager.com/gtag/js?id={{ googleAnalyticsTrackingId }}"></script>
+    <script data-testid="ga-tag-js-init">
```
colinrotherham, line 37: Because Google Tag Manager can include custom JavaScript, we can't risk putting this in `<head>`

We do add `async` here yet whilst it's non-blocking it'll still run immediately when ready

It's execution may impact lazy-loaded `async import()` scripts like the Accessible Autocomplete

Let's move it to `bodyEnd` instead and add `defer` at the very least

src/server/plugins/router.ts
alexluckett: Nice, I was speaking with Dave about something similar earlier.

`essential: true` is there a reason that one is tracked separately, given the user can't influence it (it will never be false).

colinrotherham: Yeah, sorry you missed this earlier

>essential: true is there a reason that one is tracked separately, given the user can't influence it (it will never be false)?

I've added `essential` only for completeness, it's not needed

src/server/plugins/router.ts
```diff
@@ -64,19 +65,23 @@ export default {
...
 
+          h.state('cookie_consent', decision)
           request.yar.flash('cookieConsentUpdated', true, true)
 
           return h.redirect(request.info.referrer)
+        },
+        options: {
+          validate: {
+            payload: Joi.object({
+              'cookies[additional]': Joi.string().valid('yes', 'no').required()
```
colinrotherham, line 82: Can you rename `additional` to `analytics` and same with the content?

We've got a mixture of two Design System examples at the moment

**Additional cookies**
>We’d like to set additional cookies so we can remember your settings, understand how people use the service and make improvements.

**Analytics cookies**
>We’d also like to use analytics cookies so we can understand how you use the service and make improvements.

src/server/plugins/nunjucks/context.js
```diff
@@ -28,6 +28,16 @@ export function context(request) {
   }
 
   const { params, path } = request ?? {}
+
+  let cookieConsent
+  let cookieConsentUpdated
+
+  try {
+    cookieConsent = request?.state.cookieConsent ?? undefined
+    cookieConsentUpdated = request?.yar.flash('cookieConsentUpdated').at(0)
```
colinrotherham, line 37: Can we move this preference into the consent cookie too?

Then we can easily dismiss the cookie banner client-side without needing a page reload

I've suggested [something like `hide: false`](https://github.com/DEFRA/forms-runner/pull/571/files#r1875695628) and [updating the state to match](https://github.com/DEFRA/forms-runner/pull/571/files#r1876116756)

src/typings/hapi/index.d.ts
```diff
@@ -23,6 +23,8 @@ declare module '@hapi/hapi' {
   interface Request {
     logger: Logger
     yar: Yar
+    state: object
```
colinrotherham, line 26: Do we need `state`?

src/server/plugins/router.ts
alexluckett: Sounds good. Moving this logic into a separate PR that will come later.

src/server/views/layout.html
alexluckett: Note we'll have to test this, afaik it will impact the analytics gathered. Should be fine considering the other improvements.

src/server/plugins/nunjucks/context.js
alexluckett: Will do but in a follow-up PR, thanks for the feedback. For now I'll leave it.

src/common/cookies.js
```diff
@@ -0,0 +1,53 @@
...
+export const defaultConsent = {
+  analytics: null
+}
+
+/**
+ * Parses the cookie consent policy
+ * @param {string} value
+ */
+export function parseCookieConsent(value) {
+  /** @type {CookieConsent} */
```
colinrotherham, line 13:
```suggestion
  /** @type {CookieConsent | undefined} */
```

src/common/cookies.js
```diff
@@ -0,0 +1,53 @@
...
+ * Serialises the cookie consent policy
+ * @param {CookieConsent} consent
+ * @returns {string} cookie value
+ */
+export function serialiseCookieConsent(consent) {
+  return encodeURIComponent(JSON.stringify(consent))
+}
+
+/**
+ * @param {any} consent
```
colinrotherham, line 44: Try changing this to `unknown` and we've got a bit of work to do

The `in consent` bit will throw a TypeError unless we check it's a non-null, non-array object first

```suggestion
 * @param {unknown} consent
```

<br>

> #### [`JSON.parse()` return value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#return_value)
> The [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object), [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array), string, number, boolean, or null value corresponding to the given JSON text.

src/common/cookies.js
```diff
@@ -0,0 +1,53 @@
...
+ * @returns {consent is CookieConsent}
+ */
+function isValidConsent(consent) {
+  return 'analytics' in consent
```
colinrotherham, line 48:
```suggestion
  if (consent === null || Array.isArray(consent)) {
    return false
  }

  return typeof consent === 'object' && 'analytics' in consent
```

src/server/plugins/blankie.test.ts
```diff
@@ -1,18 +1,52 @@
...
+
+    const { options } = configureBlankiePlugin()
 
     expect(options).toEqual({
       defaultSrc: ['self'],
       fontSrc: ['self', 'data:'],
       frameSrc: ['self', 'data:'],
       connectSrc: ['self'],
-      scriptSrc: ['self', 'unsafe-inline'],
+      scriptSrc: ['self', 'strict-dynamic', 'unsafe-inline'],
```
colinrotherham, line 15: This line throws CSP errors in older Safari:
https://caniuse.com/?search=strict-dynamic

All versions up to (but not including) 15.4 from March 2022 😬

Sorry, it's too much of an impact and will break the **Accessible Autocomplete** component

Could we add a new `{{ getAssetHash() }}` helper to read from the webpack JSON manifest?

src/server/views/layout.html
colinrotherham: Yeah it'll 100% affect the analytics as it'll run after our components do

That's what we want?

src/server/plugins/blankie.test.ts
colinrotherham: Older Safari just throws CSP errors, we might be OK simply because it can't understand it

But we should verify this really

davidjamesstone (approved):

src/server/plugins/nunjucks/context.js
```diff
@@ -41,6 +51,13 @@ export function context(request) {
     serviceName: config.get('serviceName'),
     serviceVersion: config.get('serviceVersion'),
     slug: params?.slug,
+    cookieConsent,
+    cookieConsentUpdated,
+    googleAnalyticsTrackingId: config.get('googleAnalyticsTrackingId'),
+    cspNonce: request?.plugins.blankie?.nonces?.script,
+    currentPath: request
+      ? `${request.path.toString()}${request.url.search}`
```
colinrotherham, line 59: Was ESLint kicking off that a string needed `.toString()`?

```suggestion
      ? `${request.path}${request.url.search}`
```

alexluckett: Leftover code from my testing, thanks!

src/server/index.ts
```diff
@@ -124,5 +125,13 @@ export async function createServer(routeConfig?: RouteConfig) {
   await server.register(pluginErrorPages)
   await server.register(blipp)
 
+  server.state('cookieConsent', {
+    ttl: 365 * 24 * 60 * 60 * 1000, // 1 year in ms
+    clearInvalid: true,
+    isSecure: config.get('isProduction'),
+    path: '/',
+    encoding: 'none' // handle this inside the application so we can share frontend/backend cookie modification
```
colinrotherham, line 133: Can probably drop `'none'` as it's [one of the defaults](https://github.com/hapijs/statehood/blob/master/lib/index.js#L45-L55)

```suggestion
    path: '/'
```

src/server/plugins/blankie.ts
```diff
@@ -1,16 +1,51 @@
 import { type ServerRegisterPluginObject } from '@hapi/hapi'
 import Blankie from 'blankie'
+import flatten from 'lodash/flatten.js'
```
colinrotherham, line 3: Mind if we just use `.flat()` in JavaScript?

I'll put an example below

src/server/plugins/blankie.ts
```diff
@@ -1,16 +1,51 @@
...
+  */
+  return {
+    plugin: Blankie,
+    options: {
+      defaultSrc: ['self'],
+      fontSrc: ['self', 'data:'],
+      connectSrc: combine(
+        ['self'],
+        gaTrackingId ? googleAnalyticsOptions.connectSrc : []
+      ),
```
colinrotherham, line 37: Example using `.flat()`

Won't need Lodash or `new Set()` for this

```suggestion
      connectSrc: [
        ['self'],
        gaTrackingId ? googleAnalyticsOptions.connectSrc : []
      ].flat(),
```

src/server/views/help/cookie-preferences.html
```diff
@@ -0,0 +1,55 @@
+{% extends 'layout.html' %}
```
colinrotherham, line 1:
```suggestion
{% extends "layout.html" %}
```

src/server/views/help/cookies.html
```diff
@@ -34,6 +39,33 @@ <h2 class="govuk-heading-m">Essential cookies</h2>
...
+        { text: "Expires" }
+      ],
+      rows: [
+        [
+          { text: "_ga" },
+          { text: "Used by Google Analytics to help us count how many people visit our forms by tracking if you’ve visited before" },
+          { text: "2 years" }
+        ],
+        [
+          { text: "_ga_" + googleAnalyticsTrackingId },
```
colinrotherham, line 60: Is this definitely correct?

The [GOV.UK Design System uses `ga_[random number]`](https://design-system.service.gov.uk/cookies/)

src/server/views/help/cookies.html
```diff
@@ -34,6 +39,33 @@ <h2 class="govuk-heading-m">Essential cookies</h2>
         ]
       ]
     }) }}
+
+    <h2 class="govuk-heading-m">Analytics cookies</h2>
+    <p class="govuk-body">We use Google Analytics cookies to collect information about how you use this form.</p>
```
colinrotherham, line 44: Just a suggestion, then we can talk separately about what we collect further down

```suggestion
    <p class="govuk-body">We use Google Analytics software to understand how people use our forms. We do this to help make sure the site is meeting the needs of its users and to help us make improvements.</p>
```

src/server/index.ts
alexluckett: I'll leave it just for the comment, makes it explict.

src/server/plugins/blankie.ts
alexluckett: I used the set to handle duplicates, in the event of a clash between GA and another dependency. After some refactoring there aren't any duplicates anymore so I'm happy to remove it.

src/server/views/help/cookies.html
alexluckett: Pretty sure it is. It's certainly not a random number, but there might be a different ID called a 'container ID' we need to look for rather than the tag ID. I'll merge this and verify in the test env.

colinrotherham (approved):

PR #572, opened 2024-12-06
Conditions exit page isPreview override validate

davidjamesstone:
This PR allows the user to bypass the page view checks while in preview mode

### DIFF
--- a/src/server/plugins/engine/models/FormModel.ts
+++ b/src/server/plugins/engine/models/FormModel.ts
@@ -297,10 +297,7 @@ export class FormModel {
       paths.push(path)
 
       // Stop at current page or with errors
-      if (
-        !isPreview &&
-        (path === currentPath || collection.getErrors(errors))
-      ) {
+      if (!isPreview && collection.getErrors(errors)) {
         break
       }
     }
--- a/src/server/plugins/engine/models/FormModel.ts
+++ b/src/server/plugins/engine/models/FormModel.ts
@@ -16,6 +16,7 @@ import { Parser, type Value } from 'expr-eval'
 import joi from 'joi'
 
 import {
+  checkFormStatus,
   findPage,
   getError,
   getPage
@@ -229,8 +230,6 @@ export class FormModel {
     state: FormState,
     options?: { validate: false }
   ): FormContext | FormContextProgress {
-    const { query } = request
-    const { force = false } = query
     const page = getPage(this, request)
 
     // Determine form paths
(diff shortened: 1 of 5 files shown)


src/server/plugins/engine/pageControllers/PageControllerBase.ts
```diff
@@ -289,7 +289,7 @@ export class PageControllerBase {
       const context = model.getFormContext(request, state)
 
       // Redirect back to last relevant page
-      if (!context.paths.includes(path)) {
+      if (!context.force && !context.paths.includes(path)) {
```
colinrotherham, line 292: Morning @davidjamesstone

Just realised that putting the override here stops us using ~`progress`~ `paths` in future

E.g. Users would be granted forced access but the back link wouldn't work correctly

Instead, can we suppress the [loop break in `getFormContext()`](https://github.com/DEFRA/forms-runner/blob/conditions-exit-page-validate-force/src/server/plugins/engine/models/FormModel.ts#L300-L303) instead?

**Before**
We stop pushing to `paths` when we hit errors
```js
// Stop at current page or with errors
if (path === currentPath || collection.getErrors(errors)) {
```

**After**
We keep pushing to `paths` when we hit errors but for form previews only
```js
// Stop at current page or with errors
if (path === currentPath || (collection.getErrors(errors) && !isPreview)) {
```

Probably don't even need to stop at the current page because it's [in the loop above](https://github.com/DEFRA/forms-runner/blob/conditions-exit-page-validate-force/src/server/plugins/engine/models/FormModel.ts#L273-L276) already

src/server/plugins/engine/models/FormModel.ts
```diff
@@ -289,13 +290,17 @@ export class FormModel {
...
     // Add paths for navigation
     for (const { collection, path } of context.relevantPages) {
       paths.push(path)
 
       // Stop at current page or with errors
-      if (path === currentPath || collection.getErrors(errors)) {
+      if (
+        !isPreview &&
+        (path === currentPath || collection.getErrors(errors))
+      ) {
```
colinrotherham, line 303: Looks ace @davidjamesstone

Can we remove the current path check mentioned in https://github.com/DEFRA/forms-runner/pull/572#discussion_r1875573566?

Think we're covered by the loop above

```suggestion
      if (!isPreview && collection.getErrors(errors)) {
```

alexluckett (approved):

PR #581, opened 2024-12-10
Move non-form properties and methods to base class

colinrotherham:
This PR moves all form-only properties, methods + getters etc to `PageController`

Makes it much easier to write base controller tests without worrying about state etc

Leaving `QuestionPageController` to do form related things

### DIFF
--- a/src/server/plugins/engine/pageControllers/PageController.test.ts
+++ b/src/server/plugins/engine/pageControllers/PageController.test.ts
@@ -24,6 +24,109 @@ describe('PageController', () => {
     controller2 = new PageController(model, page2)
   })
 
+  describe('Properties', () => {
+    it('returns path', () => {
+      expect(controller1).toHaveProperty('path', '/licence')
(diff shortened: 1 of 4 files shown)


src/server/plugins/engine/pageControllers/index.ts
```diff
@@ -1,6 +1,9 @@
 export { HomePageController } from '~/src/server/plugins/engine/pageControllers/HomePageController.js'
-export { PageController } from '~/src/server/plugins/engine/pageControllers/PageController.js'
 export { StartPageController } from '~/src/server/plugins/engine/pageControllers/StartPageController.js'
+export {
+  QuestionPageController, // Export alongside alias
+  QuestionPageController as PageController
+} from '~/src/server/plugins/engine/pageControllers/QuestionPageController.js'
```
colinrotherham, line 6: For compatibility, temporarily support `PageController` as an alias of `QuestionPageController`

PR #586, opened 2024-12-11
Match label font size/weight when amongst legends

colinrotherham:
Demo PR to show Dan

Parked for future discussions

# Before

The default GOV.UK Frontend styling uses regular weight field labels

We apply bold weight for page headings and legends as shown in [address](https://design-system.service.gov.uk/patterns/addresses/multiple/), [date](https://design-system.service.gov.uk/patterns/dates/default/) and [bank details](https://design-system.service.gov.uk/patterns/bank-details/default/) examples

Label weight before

# After

This PR shows automatic bold weight styling added to field labels when amongst legends

Label weight after


alexluckett:
Visually much nicer to look at, would love for this to be taken forward.

PR #596, opened 2024-12-17
Add Nunjucks debug macro

colinrotherham:
This has been really helpful for me, any thoughts?

Shows below each question during local development only:

Form context debug information


mokhld:
I think this can be massively helpful during development 👍 Not sure about the state keys though, but the values are self-explainable nonetheless. Are the paths in order of the traversed journey?

alexluckett:
Personally not a massive fan of changing a codebase for testing/development purposes, but I'm happy to include it if you guys find it useful.

mokhld (approved):

PR #597, opened 2024-12-17
[WIP] Exit page routing checks

colinrotherham:


src/server/plugins/engine/components/ComponentCollection.ts
```diff
@@ -232,11 +232,7 @@ export class ComponentCollection {
     return list
   }
 
-  getViewModel(
-    payload: FormPayload,
-    errors?: FormSubmissionError[],
-    conditions?: FormModel['conditions']
```
colinrotherham, line 238: We never pass in the 3rd `conditions` param so I've removed it

Over in PR https://github.com/DEFRA/forms-runner/pull/646 the 3rd param comes back as `query` to pick up `?force` direct access

src/server/plugins/engine/components/ComponentCollection.ts
```diff
@@ -250,14 +246,6 @@ export class ComponentCollection {
...
     })
 
-    if (conditions) {
-      return result.filter((item) =>
-        item.model.condition
-          ? conditions[item.model.condition]?.fn(payload)
-          : true
-      )
-    }
-
```
colinrotherham, line 260: As shown above, this is dead code

We never passed in the 3rd `conditions` param anywhere

src/server/plugins/engine/pageControllers/QuestionPageController.ts
```diff
@@ -261,10 +261,11 @@ export class QuestionPageController extends PageController {
       context: FormContext,
       h: Pick<ResponseToolkit, 'redirect' | 'view'>
     ) => {
-      const { model, viewName } = this
+      const { collection, model, viewName } = this
       const { evaluationState, state } = context
 
       const viewModel = this.getViewModel(request, context)
+      viewModel.errors = collection.getErrors(viewModel.errors)
```
colinrotherham, line 268: We already do this on POST

Without it, state errors from previous pages will appear when `?force` arrives in https://github.com/DEFRA/forms-runner/pull/646

src/server/plugins/engine/views/partials/form.html
```diff
@@ -3,7 +3,7 @@
 
 <form method="post" novalidate>
   <input type="hidden" name="crumb" value="{{ crumb }}">
-  <input type="hidden" name="action" value="continue">
+  <input type="hidden" name="action" value="validate">
```
alexluckett, line 6: Do we need a separate `validate` action, as opposed to just `continue`? I can't find a reference to `continue` being used anywhere now.

src/server/plugins/engine/pageControllers/QuestionPageController.ts
```diff
@@ -241,6 +238,14 @@ export class QuestionPageController extends PageController {
     return cacheService.getState(request)
   }
 
+  async setState(
+    request: FormRequest | FormRequestPayload,
+    state: FormSubmissionState
+  ) {
+    const { cacheService } = request.services([])
+    return cacheService.setState(request, state)
+  }
```
alexluckett, line 247: This feels like an unneccessary wrapper, especially given it's only used once.

src/server/plugins/engine/pageControllers/QuestionPageController.ts
```diff
@@ -398,68 +384,32 @@ export class QuestionPageController extends PageController {
...
-        // This is required to ensure we don't navigate
-        // to an incorrect page based on stale state values
-        this.getNextPath(
-          model.getFormContext(request, state, {
-            validate: false
-          })
-        )
-      )
+      // Save and proceed
+      await this.setState(request, state)
```
alexluckett, line 408: Can you clarify the reason for swapping `mergeState` with `setState`? Presumably it's because the state is fully formed and has been modified, so merging is an unneccessary operation?

src/server/plugins/engine/views/partials/form.html
colinrotherham: Yeah I'll explain

I've used `FormAction.Continue` when the `payload` is valid but not necessarily the `state`

E.g. When adding your first pizza, you don't want `min: 25` to shout at you for every POST

https://github.com/DEFRA/forms-runner/blob/0f94f61855da59533d1209242688f4fbe835924c/src/server/plugins/engine/views/repeat-list-summary.html#L34

src/server/plugins/engine/pageControllers/QuestionPageController.ts
colinrotherham: Yeah it is, except it's `mergeState()` that becomes redundant

Have a look at commit https://github.com/DEFRA/forms-runner/pull/646/commits/98de460de165ece6f8fab3bd6d99a467d0ce6ab3 in [PR #646](https://github.com/DEFRA/forms-runner/pull/646) where we render the merged-but-not-saved-yet state

Want me to bring those changes into this PR?

src/server/plugins/engine/pageControllers/QuestionPageController.ts
colinrotherham: Yeah, we're grabbing `const { state } = context` further up

It would be like doing this:

```ts
await this.mergeState(request, context.state, context.state)
```

Previously we used to re-read the state before every write until this PR:

* https://github.com/DEFRA/forms-runner/pull/599

src/server/plugins/engine/pageControllers/QuestionPageController.ts
alexluckett: No, that's fine now I have the context. For next time I'd move this kind of change into #646 rather than bringing those changes forward.

src/server/plugins/engine/views/partials/form.html
alexluckett: Got it, makes sense. I wonder if we can bake that intent into the names instead? Maybe `validate` -> `commit`?

colinrotherham: Is it a blocker? Neither did "Continue" say it was going to validate or save the answers

alexluckett (approved):

PR #599, opened 2024-12-18
Exit pages: Reduce controller state read/write

colinrotherham:
Reduces Redis read/write calls by sharing the existing state where possible

Flagged during the ["exit page bug" #482470](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/482470) fixes

### DIFF
--- a/src/server/services/cacheService.ts
+++ b/src/server/services/cacheService.ts
@@ -35,14 +35,23 @@ export class CacheService {
     return cached || {}
   }
 
-  async mergeState(request: FormRequest | FormRequestPayload, value: object) {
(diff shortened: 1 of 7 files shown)


src/server/services/cacheService.ts
```diff
@@ -1,5 +1,5 @@
 import { type Request, type Server } from '@hapi/hapi'
-import { merge } from '@hapi/hoek'
+import * as Hoek from '@hapi/hoek'
```
mokhld, line 2: TIL

mokhld (approved):

PR #614, opened 2024-12-24
Only redirect to new repeater item when list is empty

colinrotherham:
This PR updates the handling of the repeater page

Fixes [bug #490111](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490111)

**Before**
When visiting a repeater page, we redirect to the new item page

**After**
When visiting a repeater page:

* Empty repeater state is redirecting to the new item page
* Populated repeater state is redirected to the mini summary page with "Add another" or "Continue"


colinrotherham:
Wrong branch name, going to re-open this

PR #619, opened 2024-12-24
Add confirm delete page to file upload

colinrotherham:
The file upload "Remove" button's `style="margin-top: -7px"` attribute has been blocked since:

* https://github.com/DEFRA/forms-runner/pull/571

As an alternative fix I've shared the confirm delete page from **Add another** with repeaters

This resolves [bug #490120](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490120)

# Remove link

This PR uses the [`govukSummaryList()` macro](https://design-system.service.gov.uk/components/summary-list/) to render a "Remove" link with confirmation page

Remove link

# Confirmation page

Confirm remove


src/server/plugins/engine/pageControllers/FileUploadPageController.ts
```diff
@@ -131,7 +132,62 @@ export class FileUploadPageController extends QuestionPageController {
...
+
+      const state = await this.getState(request)
+      const uploadState = state.upload?.[path] ?? { files: [] }
+
+      const fileToRemove = uploadState.files.find(
+        (file) => file.uploadId === params.itemId
+      )
+
+      if (!fileToRemove) {
+        return Boom.notFound('File to delete not found')
```
mokhld, line 151: Would be great to place these error messages in an enum so we only have one place to change them

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
```diff
@@ -141,13 +197,28 @@ export class FileUploadPageController extends QuestionPageController {
       const state = await this.getState(request)
       const uploadState = state.upload?.[path] ?? { files: [] }
 
-      // Check for any removed files in the POST payload
-      const removed = await this.checkRemovedFiles(request, uploadState)
+      const { confirm } = this.getFormData(request)
 
-      if (removed) {
+      // Check for any removed files in the POST payload
+      if (confirm === true) {
```
mokhld, line 203: Is the === true comparison necessary? Since confirm is already a boolean, the if block would execute if it is truthy

src/server/plugins/engine/plugin.ts
```diff
@@ -408,25 +409,30 @@ export const plugin = {
...
       const { app, params } = request
       const page = getPage(app.model, request)
 
-      if (!(page instanceof RepeatPageController)) {
-        throw Boom.notFound(`No repeater page found for /${params.path}`)
+      if (
+        !(
+          page instanceof RepeatPageController ||
+          page instanceof FileUploadPageController
+        )
```
mokhld, line 424: 👌

mokhld (approved):
Nice work @colinrotherham Just a small question on the truthy comparison, but not a biggie at all.

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
colinrotherham: Thanks @mokhld, good point

The schema is `Joi.boolean().default(false)` so I'll update **RepeatPageController.ts** too

https://github.com/DEFRA/forms-runner/blob/c6c545ac8d9ae90b4bd57d69580f1a8eca382bf6/src/server/plugins/engine/pageControllers/RepeatPageController.ts#L339

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
colinrotherham: Mind if we start a separate discussion on this?

Need a way to support template strings too in lots of our error messages, but sounds good still

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
colinrotherham: Fixed in https://github.com/DEFRA/forms-runner/pull/619/commits/a3741086ae612cedfebc512bc274537e3d676c5c

PR #620, opened 2024-12-24
Check for maximum repeater state on continue

colinrotherham:
This PR includes 2x changes to fix [bug #490108](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490108)

1. **Fix state validation to include non-field error keys**
We currently filter out the `stateSchema` repeater key

2. **Check for maximum repeater state on continue**
We now check for maximum repeater items on continue (not just add another)

Maximum repeater items

☝️ The ability to add more than the maximum items will be addressed in [bug #490107](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490107)

### DIFF
--- a/src/server/plugins/engine/models/FormModel.ts
+++ b/src/server/plugins/engine/models/FormModel.ts
@@ -284,11 +284,16 @@ export class FormModel {
     const { isPreview } = checkFormStatus(request.path)
 
     // Add paths for navigation
-    for (const { collection, path } of context.relevantPages) {
+    for (const { keys, path } of context.relevantPages) {
       paths.push(path)
 
-      // Stop at current page or with errors
-      if (!isPreview && collection.getErrors(errors)) {
+      // Stop at page with errors
+      if (
+        !isPreview &&
+        errors?.some(({ name, path }) => {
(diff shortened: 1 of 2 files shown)


mokhld (approved):

src/server/plugins/engine/models/FormModel.ts
```diff
@@ -284,11 +284,16 @@ export class FormModel {
...
+    for (const { keys, path } of context.relevantPages) {
       paths.push(path)
 
-      // Stop at current page or with errors
-      if (!isPreview && collection.getErrors(errors)) {
+      // Stop at page with errors
+      if (
+        !isPreview &&
+        errors?.some(({ name, path }) => {
+          return keys.includes(name) || keys.some((key) => path.includes(key))
```
colinrotherham, line 294: Our helper `collection.getErrors()` only included errors from collection fields

State schema errors for the repeater `options.name` nested value were excluded:

https://github.com/DEFRA/forms-runner/blob/c1fecf5f43ffd233a9618005c80eb4d9b7a959c7/src/server/plugins/engine/pageControllers/RepeatPageController.ts#L44-L52

PR #627, opened 2024-12-30
Use Joi to validate form params object

colinrotherham:
This PR renames page `getFormData()` to `getFormParams()` and adds Joi validation

For clarity, this prevents confusion with `getFormDataFromState()`

```patch
- const { confirm } = this.getFormData(request)
+ const { confirm } = this.getFormParams(request)
  
  // Check for any removed files in the POST payload
  if (confirm) {
```

### DIFF
--- a/src/server/plugins/engine/types.ts
+++ b/src/server/plugins/engine/types.ts
@@ -71,19 +71,18 @@ export interface FormSubmissionError
   text: string // e.g: 'Date field must be a real date'
 }
 
+export interface FormParams {
+  action?: FormAction
+  confirm?: true
+  crumb?: string
+  itemId?: string
+}
+
 /**
  * Form POST for question pages
  * (after Joi has converted value types)
  */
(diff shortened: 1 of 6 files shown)


src/server/plugins/engine/types.ts
```diff
@@ -71,19 +71,18 @@ export interface FormSubmissionError
   text: string // e.g: 'Date field must be a real date'
 }
 
+export interface FormParams {
+  action?: FormAction
+  confirm?: true
```
mokhld, line 76: @colinrotherham should this be `boolean`?

colinrotherham: Glad you spotted this

Remember the `confirm === true` PR feedback the other day?

You can't submit `confirm=false` so it's only ever `true` or `undefined` in the HTML

mokhld: aha 💡

mokhld (approved):

PR #629, opened 2024-12-31
Fix repeater state being overwritten when switching page type

colinrotherham:
This PR adds `getFormDataFromState()` to the "Add another" page type

We can now remove `getState()`, `setRepeatAppData()` and `getRepeatAppData()`

Fixes [bug #490113](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490113)


mokhld (approved):
Nice, much cleaner now 👌

PR #637, opened 2025-01-02
Automatically generate back links

colinrotherham:
This PR removes the `progress[]` array and uses `context.paths` to calculate the back link

Fixes [bug #490505](https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/490505)

### DIFF
--- a/src/server/plugins/engine/pageControllers/QuestionPageController.ts
+++ b/src/server/plugins/engine/pageControllers/QuestionPageController.ts
@@ -2,6 +2,7 @@ import {
   ComponentType,
   hasComponents,
   hasNext,
+  hasRepeater,
   type Link,
   type Page
 } from '@defra/forms-model'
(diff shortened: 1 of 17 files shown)


src/server/plugins/engine/pageControllers/QuestionPageController.ts
```diff
@@ -345,40 +349,45 @@ export class QuestionPageController extends PageController {
...
 
-  /**
-   * Get the back link for a given progress.
-   */
-  protected getBackLink(progress: string[]) {
-    return progress.at(-2)
+    // No back link
+    if (!backPath) {
+      return
+    }
```
alexluckett, line 384: This logic feels like it'd sit better in the relevant page's controller, rather than the generic question page controller, otherwise a change in one controller will impact this one. 

Not a blocker for now but a goal to move towards.

alexluckett (approved):

PR #669, opened 2025-01-15
feature/485768 page conditions

davidjamesstone:


alexluckett (commented):
Looking good. If this gets more complicated, we should look at splitting out our v1/v2 engine logic into separate encapsulated functions and using something like the strategy pattern to switch between the logic for engine v1/v2 depending on the form definition's engine version.

e.g.

**engineStrategies.js**
```
class EngineStrategyV1 {
    getNextPath() { }
    // anything else we need
}

class EngineStrategyV2 {
    getNextPath() { }
    // anything else we need
}
```

**QuestionPageController.ts**
```
getNextPath(context: FormContext) {
    const { model, next, path } = this
    const { evaluationState } = context
    const summaryPath = this.getSummaryPath()
    const statusPath = this.getStatusPath()

    // Walk from summary page (no next links) to status page
    let defaultPath = path === summaryPath ? statusPath : undefined

    return engineStrategy.getNextPath() ?? defaultPath // engineStrategy would be passed in ideally as a dependency
}

questionPageController = new QuestionPageController(engineStrategy) // engineStrategy is either EngineStrategyV1 or EngineStrategyV2 based on model.engine
```

alexluckett (approved):

src/server/plugins/engine/views/index.html
```diff
@@ -24,7 +24,7 @@
       {% include "partials/heading.html" %}
 
       {% block form %}
-        {% if page.next | length %}
+        {% if page.allowContinue %}
```
alexluckett, line 27: Nice 👌

alexluckett (approved):

PR #676, opened 2025-01-23
fix: prevent 500s from malformed requests in crumb generation

mokhld:
**Summary of Attempted Solutions That Didn't Work:**

1. **Yar Configuration**
   - Tried enabling `storeBlank: true` in session plugin
   - Didn't resolve the state initialisation timing issue and there's the issue of it breaking our auth tests mentioned in the comment

3. **Crumb Plugin Changes**
   - Added `skip` function with state checks
   - Added try/catch and defensive checks
   - Still hit errors before skip function was called

**Root Issue:**
The error occurs because crumb tries to access `request.state` before Yar has a chance to initialise it

### DIFF
--- a/src/server/plugins/nunjucks/context.js
+++ b/src/server/plugins/nunjucks/context.js
@@ -9,7 +9,10 @@ import { parseCookieConsent } from '~/src/common/cookies.js'
 import { config } from '~/src/config/index.js'
 import { createLogger } from '~/src/server/common/helpers/logging/logger.js'
 import { PREVIEW_PATH_PREFIX } from '~/src/server/constants.js'
-import { encodeUrl } from '~/src/server/plugins/engine/helpers.js'
+import {
+  encodeUrl,
+  safeGenerateCrumb
+} from '~/src/server/plugins/engine/helpers.js'
 
 const logger = createLogger()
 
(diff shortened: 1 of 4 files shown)


src/server/plugins/nunjucks/context.js
```diff
@@ -54,7 +54,14 @@ export function context(request) {
       serviceName: config.get('serviceName'),
       serviceVersion: config.get('serviceVersion')
     },
-    crumb: request?.server.plugins.crumb.generate?.(request),
+    // only generate crumb if plugin exists and is enabled for this route
+    crumb:
+      request?.server.plugins.crumb.generate &&
+      request.route.settings.plugins?.crumb !== false &&
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
mokhld, line 61: I  attempted to mark as optional in Hapi type definitions but it can't seem be overridden, and can be undefined for malformed requests

mokhld: We could also make this more defensive if needed but it's scalability is questionable e.g. 
```
 request?.state &&
!request.path.includes('//')
```

alexluckett (approved):
Thank you 🙌

PR #693, opened 2025-02-17
Feature 478377 : File Upload Error Handling

davidjamesstone:
## Ticket
[https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/478377](url)

## Description
This PR introduces file upload error handling and enhancements using the existing code rather than relying on Javascript. Another PR will follow to provide the remaining UX refinements.

### DIFF
--- a/src/server/plugins/engine/helpers.ts
+++ b/src/server/plugins/engine/helpers.ts
@@ -266,11 +266,14 @@ export function safeGenerateCrumb(
 }
 
 /**
- * Calculates an exponential backoff delay (in milliseconds) based on the current depth,
- * but capping the delay at 60,000 milliseconds (60 seconds).
+ * Calculates an exponential backoff delay (in milliseconds) based on the current retry depth,
+ * using a base delay of 2000ms (2 seconds) and doubling for each additional depth, while capping the delay at 30,000ms (30 seconds).
  * @param depth - The current retry depth (1, 2, 3, …)
- * @returns The calculated delay in milliseconds
+ * @returns The calculated delay in milliseconds.
  */
 export function getExponentialBackoffDelay(depth: number): number {
-  return Math.min(2 ** depth * 1000, 60000)
+  const BASE_DELAY_MS = 2000 // 2 seconds initial delay
+  const CAP_DELAY_MS = 30000 // cap each delay to 30 seconds
+  const delay = BASE_DELAY_MS * 2 ** (depth - 1)
+  return Math.min(delay, CAP_DELAY_MS)
 }
--- a/src/server/plugins/engine/helpers.ts
+++ b/src/server/plugins/engine/helpers.ts
@@ -264,3 +264,13 @@ export function safeGenerateCrumb(
 
   return request.server.plugins.crumb.generate(request)
 }
+
+/**
+ * Calculates an exponential backoff delay (in milliseconds) based on the current depth,
+ * but capping the delay at 60,000 milliseconds (60 seconds).
+ * @param depth - The current retry depth (1, 2, 3, …)
+ * @returns The calculated delay in milliseconds
+ */
+export function getExponentialBackoffDelay(depth: number): number {
+  return Math.min(2 ** depth * 1000, 60000)
+}
--- a/src/server/plugins/engine/pageControllers/FileUploadPageController.ts
+++ b/src/server/plugins/engine/pageControllers/FileUploadPageController.ts
@@ -310,7 +310,7 @@ export class FileUploadPageController extends QuestionPageController {
       } else if (statusResponse.uploadStatus === UploadStatus.pending) {
         if (depth > 6) {
           throw Boom.gatewayTimeout(
-            `Giving up waiting for getUploadStatus for ${uploadId} to complete`
+            `Giving up waiting for ${uploadId} to complete`
           )
         }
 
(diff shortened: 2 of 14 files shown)


src/server/plugins/engine/helpers.ts
```diff
@@ -264,3 +264,13 @@ export function safeGenerateCrumb(
...
 }
+
+/**
+ * Calculates an exponential backoff delay (in milliseconds) based on the current depth,
+ * but capping the delay at 60,000 milliseconds (60 seconds).
+ * @param depth - The current retry depth (1, 2, 3, …)
+ * @returns The calculated delay in milliseconds
+ */
+export function getExponentialBackoffDelay(depth: number): number {
+  return Math.min(2 ** depth * 1000, 60000)
```
whitewaterdesign, line 275: wondering, would having the cap as an environment variable be helpful?

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
```diff
@@ -290,108 +285,83 @@ export class FileUploadPageController extends QuestionPageController {
...
-      for (let index = 0; index < indexes.length; index++) {
-        const idx = indexes[index]
-        const result = results[index]
+      const delay = getExponentialBackoffDelay(depth)
+      request.logger.info(
+        `Waiting ${delay / 1000} seconds for ${uploadId} to complete`
+      )
+      await wait(delay)
+      return this.checkUploadStatus(request, state, depth + 1)
+    }
```
whitewaterdesign, line 329: Love some recursion 😍

src/server/services/cacheService.ts
```diff
@@ -75,6 +76,26 @@ export class CacheService {
...
+
+  setFlash(
+    request: FormRequest | FormRequestPayload,
+    message: { errors: FormSubmissionError[] }
+  ) {
+    const key = this.Key(request)
+
+    request.yar.flash(key.id, message)
+  }
+
```
whitewaterdesign, line 98: like this. 👍 great pattern

whitewaterdesign (approved):
Nice

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
```diff
@@ -290,108 +285,83 @@ export class FileUploadPageController extends QuestionPageController {
...
-
-      // Update state with the latest result
-      for (let index = 0; index < indexes.length; index++) {
-        const idx = indexes[index]
-        const result = results[index]
+      const delay = getExponentialBackoffDelay(depth)
+      request.logger.info(
+        `Waiting ${delay / 1000} seconds for ${uploadId} to complete`
+      )
+      await wait(delay)
```
jbarnsley10, line 327: Am I understanding this correctly? If you wait 2 secs, then wait for 4 secs, then wait for 8 etc up to 60 seconds, you'll be waiting for 2 + 4 + 8 + 16 + 32 + 60. And that's more than the 60 seconds CDP timeout. Is it not better to avoid exponential back-off and just do a linear say 10 secs per recursion (up to 6 times)?

jbarnsley10 (approved):
Looks great. Just that one question about total waiting time.

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
whitewaterdesign: Yeah, `getExponentialBackoffDelay` handles this.

whitewaterdesign: Yeah, think @jbarnsley10 is right - the max fn only sets the max per depth, not the max per entire sum with the previous waits included - 2+4+8+16+30 = 60

whitewaterdesign: Yeah, think @jbarnsley10 is right on this one.  By the time the backoff delay gets to 30 it will already hit the timeout, as with the delays 60 seconds will have already passed.  The actual backoff delay needs to be 2 * the timeout as the rule is that the sum of the powers of 2 up to 2n−12n−1 is exactly 2n−12n−1 (since we are missing the 1, it's exactly 2^n)

whitewaterdesign: $`\sum_{n=1}^{\infty} 2^{n}`$

src/server/plugins/engine/helpers.ts
whitewaterdesign: Setting 60000 to 60000 / 2 would resolve Jez's point.

whitewaterdesign: or even (60000 / 2) - 2000 or something like that

src/server/plugins/engine/pageControllers/FileUploadPageController.ts
mokhld: Thanks guys for your input. I've refactored the retry logic. Once the cumulative exponential backoff implies roughly 60 seconds of waiting (i.e. at a depth of 5 or more), we immediately reinitialise the upload and throw a `Boom.gatewayTimeout()` error. Thus we avoid exceeding the 60-second CDP timeout whilst keeping the code relatively low-complexity.

whitewaterdesign (approved):
Looks good to me 👍

PR #713, opened 2025-02-25
[EXAMPLE] Intercept file upload and reload when status is ready

alexluckett:


alexluckett:
@mokhld an example of what I meant. I copied your previous API from your last PR. Feel free to bin this, I haven't even executed it - no idea if it'll work.

PR #714, opened 2025-02-27
Feature 478377: File-upload JS file submit and polling

mokhld:
Done as part of https://dev.azure.com/defragovuk/DEFRA-CDP/_boards/board/t/DXT%20Team/Stories?System.IterationPath=%40currentIteration&workitem=478377 and a follow-up to https://github.com/DEFRA/forms-runner/pull/710.

This introduces the file submission directly from the browser with polling of the CDP Uploader file scan status. After receiving a successful (`ready`) response from CDP, we then refresh the page. By this time, the CDP callback to our service would've already ingested the response and updated the state.

### DIFF
--- a/src/client/javascripts/file-upload.js
+++ b/src/client/javascripts/file-upload.js
@@ -253,7 +253,6 @@ function handleStandardFormSubmission(
  * @param {HTMLFormElement} formElement - The form element
  * @param {HTMLInputElement} fileInput - The file input element
  * @param {HTMLButtonElement} uploadButton - The upload button
- * @param {File | null} selectedFile - The selected file
  * @param {HTMLElement | null} errorSummary - The error summary container
  * @param {string | undefined} uploadId - The upload ID
  * @returns {boolean} Whether the event was handled
@@ -263,7 +262,6 @@ function handleAjaxFormSubmission(
   formElement,
   fileInput,
   uploadButton,
-  selectedFile,
   errorSummary,
   uploadId
 ) {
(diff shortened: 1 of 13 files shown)


alexluckett (approved):
Would be good to have this timeout after a max duration (e.g. if no 'ready' file arrives after 10 mins). Given the urgency of this PR we can do that as a follow-up and let Jignesh start testing.

mokhld:
> Would be good to have this timeout after a max duration (e.g. if no 'ready' file arrives after 10 mins). Given the urgency of this PR we can do that as a follow-up and let Jignesh start testing.

Yeah agreed, I added it in to limit to 5 mins which seems reasonable.

src/client/javascripts/file-upload.js
```diff
@@ -228,15 +349,26 @@ export function initFileUpload() {
...
+        event,
+        formElement,
+        fileInput,
+        uploadButton,
+        selectedFile,
+        /** @type {HTMLElement | null} */ (errorSummary),
+        uploadId
+      )
+    ) {
+      return
```
alexluckett, line 364: rather than returning here, could we just call `handleStandardFormSubmission` first and then perform `handleAjaxFormSubmission`? Looks like the content of the standard submission is duplicated in the ajax version anyway - makes sense to use both?

src/client/javascripts/file-upload.js
```diff
@@ -183,24 +185,143 @@ function asHTMLElement(element) {
...
+  event,
+  formElement,
+  fileInput,
+  uploadButton,
+  selectedFile,
+  errorSummary,
+  uploadId
+) {
+  if (!formElement.action || !uploadId) {
+    return false
```
alexluckett, line 271: tests seem to be immediately hitting this and not going on to test the code below it

alexluckett (approved):

PR #719, opened 2025-02-28
Adjust the label weighting when multiple form components are on a page

davidjamesstone:


src/server/plugins/engine/pageControllers/QuestionPageController.ts
```diff
@@ -151,6 +151,17 @@ export class QuestionPageController extends PageController {
       }
 
       showTitle = !isPageHeading
+    } else if (formComponents.length > 1) {
+      // When there is more than one form component,
+      // adjust the label/legends to give equal prominence
+      for (const { model } of formComponents) {
+        if (model.fieldset?.legend) {
+          model.fieldset.legend.classes = 'govuk-fieldset__legend--m'
```
jbarnsley10, line 159: Should these be appending the class in case there are already classes set for the legend/label ?

davidjamesstone: I wondered the same but user's can't apply custom classes to either the label or legend. If "classes" is passed as a `component.options`, it get applied to the root of the component there's so no risk of us overwriting

jbarnsley10 (approved):
LGTM

PR #722, opened 2025-02-28
Fix: Adjust fetch options for file uploads in local development

mokhld:
The PR modifies fetch to:

- Use `redirect: 'follow'` instead of `'manual'` when a proxy URL is present locally as in production, we're receiving the redirect from CDP but not doing anything with it.
- Add `mode: 'no-cors'` to handle CORS in local development.

Done on the back off: https://github.com/DEFRA/forms-runner/pull/720/files

### DIFF
--- a/src/client/javascripts/file-upload.js
+++ b/src/client/javascripts/file-upload.js
@@ -272,13 +272,21 @@ function handleAjaxFormSubmission(
   event.preventDefault()
 
   const formData = new FormData(formElement)
+  const isLocalDev = !!formElement.dataset.proxyUrl
   const uploadUrl = formElement.dataset.proxyUrl ?? formElement.action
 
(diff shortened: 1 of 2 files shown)


davidjamesstone (approved):

alexluckett:
Does this mean for local development, it won't poll (as it will follow the path) in the frontend but do a long wait?

mokhld:
> Does this mean for local development, it won't poll (as it will follow the path) in the frontend but do a long wait?
Nope, it will still poll as standard if JS is enabled locally as the polling is separate from the redirect behaviour. Locally, the proxy will return a redirect that we'll action with `follow` and `mode: 'no-cors'` to handle the redirect and avoid the CORS error that we see, whilst in prod, we essentially ignore the redirect sent from CDP.

PR #740, opened 2025-03-07
512731_Add cacheing for static assets

nayij:
Add cache for static assets  - as highlighted during performance testing.
Not sure of any side effect of this though.
Screenshot 2025-03-07 at 23 40 19


nayij:
Feedback: Avoid caching for local development.

davidjamesstone (approved):

PR #744, opened 2025-03-11
fix: Support field names with special characters in condition express…

mokhld:
…ions


alexluckett:
I'm guessing this is superceded by https://github.com/DEFRA/forms-designer/pull/745?

PR #759, opened 2025-04-02
Allow more than 1 error message for UK Address Field components

davidjamesstone:
https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/530750

### DIFF
--- a/src/server/plugins/engine/components/ComponentCollection.ts
+++ b/src/server/plugins/engine/components/ComponentCollection.ts
@@ -1,4 +1,4 @@
-import { type ComponentDef } from '@defra/forms-model'
+import { ComponentType, type ComponentDef } from '@defra/forms-model'
 import joi, {
   type CustomValidator,
   type ErrorReportCollection,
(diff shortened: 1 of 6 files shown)


jbarnsley10 (commented):
Looks good. Would be good to get multiple errors per line in the unit test to prove it only displays a single error per address component field, but I presume you've proved that locally anyway

davidjamesstone:
> Looks good. Would be good to get multiple errors per line in the unit test to prove it only displays a single error per address component field, but I presume you've proved that locally anyway

I wanted to do this but I can't actually think of a scenario that there would be more than 1 error per field.

## Errors scenarios
Address line 1: Missing or over 100 chars
Address line 2: Over 100 chars
Town: Missing or over 100 chars
Postcode: Missing or regex pattern

So I added the code to limit to 1 error per field as a back stop, but I can't actually see how it'd ever be more than one. I did test it locally but only through forcing another error into the list manually.

alexluckett (approved):

PR #760, opened 2025-04-02
[DRAFT] Refactor: migrate to forms-engine-plugin

alexluckett:
Don't panic! This is a playground for me to experience some of the pain points of using the plugin.


alexluckett:
Abandoning this in favour of the approach in https://github.com/DEFRA/forms-runner/pull/780

PR #773, opened 2025-04-14
Fix: Remove aria-describedby attribute based on the presence of an error summary title element 

mokhld:
This PR enhances the accessibility in file upload component by conditionally setting or removing `aria-describedby` attribute based on the presence of an error summary title element. Added unit tests to verify the new behaviour for both scenarios.


jbarnsley10 (approved):
LGTM

PR #774, opened 2025-04-14
Feat/526386  render short description

whitewaterdesign:
## Error message uses short description if it exists

The error message should now use the short description field if it exists. An example:

image

This completes the following ticket:

https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/526386

A new label field now exists on ComponentBase and this is used rather than title.  Slightly more complex logic for Date fields.

### DIFF
--- a/src/server/plugins/engine/components/AutocompleteField.test.ts
+++ b/src/server/plugins/engine/components/AutocompleteField.test.ts
@@ -317,9 +317,8 @@ describe.each([
             shortDescription: 'Your example text',
             name: 'myComponent',
             type: ComponentType.AutocompleteField,
-
-            options: {},
-            schema: {}
+            list: 'ABCE',
+            options: {}
           } satisfies AutocompleteFieldComponent,
           assertions: [
             {
--- a/src/server/plugins/engine/components/AutocompleteField.test.ts
+++ b/src/server/plugins/engine/components/AutocompleteField.test.ts
@@ -34,7 +34,7 @@ describe.each([
       list: listString,
       examples: listStringExamples,
       allow: ['1', '2', '3', '4'],
-      shortDescription: 'my string list'
+      shortDescription: 'My string list'
     }
   },
   {
@@ -50,7 +50,7 @@ describe.each([
       list: listNumber,
       examples: listNumberExamples,
       allow: [1, 2, 3, 4],
-      shortDescription: 'my number list'
+      shortDescription: 'My number list'
     }
   }
 ])('AutocompleteField: $component.title', ({ component: def, options }) => {
@@ -156,7 +156,7 @@ describe.each([
 
         expect(result.errors).toEqual([
           expect.objectContaining({
-            text: `Enter ${def.title.toLowerCase()}`
+            text: `Enter ${lowerFirst(def.title)}`
           })
         ])
       })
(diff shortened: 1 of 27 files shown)


src/server/plugins/engine/components/AutocompleteField.test.ts
```diff
@@ -32,7 +33,8 @@ describe.each([
     options: {
       list: listString,
       examples: listStringExamples,
-      allow: ['1', '2', '3', '4']
+      allow: ['1', '2', '3', '4'],
+      shortDescription: 'my string list'
```
jbarnsley10, line 37: To avoid confusion, probably a good idea to start every short description with an uppercase letter

src/server/plugins/engine/components/ComponentBase.ts
```diff
@@ -43,6 +44,10 @@ export class ComponentBase {
     this.type = def.type
     this.name = def.name
     this.title = def.title
+    this.label =
```
jbarnsley10, line 47: Would be good to adda test for the different scenarios here

src/server/plugins/engine/components/CheckboxesField.test.ts
```diff
@@ -172,7 +194,7 @@ describe.each([
 
         expect(result.errors).toEqual([
           expect.objectContaining({
-            text: `Select ${def.title.toLowerCase()}`
+            text: `Select ${options.label.toLowerCase()}`
```
davidjamesstone, line 197: These should probably be changed to use `lowerFirst` rather than `toLowerCase`, WDYT?

src/server/plugins/engine/components/ComponentBase.ts
davidjamesstone, line 47: Will non-form components support short description? If not, maybe the place for this code is `FormComponent.ts`

davidjamesstone (approved):

jbarnsley10 (approved):
Looks great. Nice and simple. Would be good to cover off the shortDesc possible combinations in a unit test

src/server/plugins/engine/components/ComponentBase.ts
alexluckett: I can't see a reason a non-input would need it, no.

jbarnsley10: Yep, that's a good point Dave. FormComponent is probably the best place

whitewaterdesign: there should be tests in every component - I think there isn't a test for ComponentBase

src/server/plugins/engine/components/CheckboxesField.test.ts
whitewaterdesign: Yeah, you're right, I should have done that.  That was pure laziness.

alexluckett: We have very different definitions of laziness.

src/server/plugins/engine/components/ComponentBase.ts
whitewaterdesign: I've changed the line now so that there's full coverage.

whitewaterdesign: Thanks, all done.

PR #778, opened 2025-04-16
Enhanced error messages for MultilineTextField and YesNoField

jbarnsley10:
Added file upload errors messages
Relies on file type definitions in forms-designer so will error while building until designer is merged and forms-model version updated here

### DIFF
--- a/src/server/plugins/engine/components/FileUploadField.ts
+++ b/src/server/plugins/engine/components/FileUploadField.ts
@@ -272,7 +272,20 @@ export class FileUploadField extends FormComponent {
   getAllPossibleErrors(): ErrorMessageTemplateList {
     return {
       baseErrors: [
-        { type: 'selectRequired', template: messageTemplate.selectRequired }
+        { type: 'selectRequired', template: messageTemplate.selectRequired },
+        {
+          type: 'filesMimes',
+          template: 'The selected file must be a {{#limit}}'
+        },
+        {
+          type: 'filesSize',
+          template: 'The selected file must be smaller than 100MB'
+        },
+        { type: 'filesEmpty', template: 'The selected file is empty' },
+        {
+          type: 'filesError',
+          template: 'The selected file could not be uploaded – try again'
+        }
       ],
       advancedSettingsErrors: [
         {
--- a/src/server/plugins/engine/components/ListFormComponent.ts
+++ b/src/server/plugins/engine/components/ListFormComponent.ts
@@ -1,4 +1,5 @@
 import {
+  ComponentType,
   type Item,
   type List,
   type ListComponentsDef,
@@ -21,6 +22,7 @@ import {
   type FormSubmissionError,
   type FormSubmissionState
 } from '~/src/server/plugins/engine/types.js'
+import { convertToLanguageMessages } from '~/src/server/utils/type-utils.js'
 
 export class ListFormComponent extends FormComponent {
   declare options: Extract<
@@ -78,6 +80,12 @@ export class ListFormComponent extends FormComponent {
 
     if (options.customValidationMessages) {
       formSchema = formSchema.messages(options.customValidationMessages)
+    } else if (def.type === ComponentType.YesNoField) {
+      formSchema = formSchema.messages(
+        convertToLanguageMessages({
+          'any.required': messageTemplate.selectYesNoRequired
+        })
+      )
     }
 
     this.formSchema = formSchema
--- a/src/server/plugins/engine/components/MultilineTextField.ts
+++ b/src/server/plugins/engine/components/MultilineTextField.ts
@@ -73,6 +73,13 @@ export class MultilineTextField extends FormComponent {
       })
     } else if (options.customValidationMessages) {
       formSchema = formSchema.messages(options.customValidationMessages)
+    } else if (
+      typeof schema.max === 'number' &&
+      typeof schema.min === 'number'
+    ) {
+      const minMaxErrorText = this.buildMinMaxText(schema.min, schema.max)
+      formSchema = formSchema.min(schema.min).message(minMaxErrorText)
+      formSchema = formSchema.max(schema.max).message(minMaxErrorText)
     }
 
     this.formSchema = formSchema.default('')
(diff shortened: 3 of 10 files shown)


src/server/plugins/engine/components/FileUploadField.ts
```diff
@@ -272,7 +272,20 @@ export class FileUploadField extends FormComponent {
   getAllPossibleErrors(): ErrorMessageTemplateList {
     return {
       baseErrors: [
-        { type: 'selectRequired', template: messageTemplate.selectRequired }
+        { type: 'selectRequired', template: messageTemplate.selectRequired },
+        {
+          type: 'filesMimes',
+          template: 'The selected file must be a {{#limit}}'
```
davidjamesstone, line 278: Should this be {{#limit}}?

src/server/plugins/engine/components/ListFormComponent.ts
```diff
@@ -78,6 +80,12 @@ export class ListFormComponent extends FormComponent {
 
     if (options.customValidationMessages) {
       formSchema = formSchema.messages(options.customValidationMessages)
+    } else if (def.type === ComponentType.YesNoField) {
+      formSchema = formSchema.messages(
```
davidjamesstone, line 84: Feels like this should live in `YesNoField.ts`

src/server/plugins/error-preview/error-preview-helper.js
```diff
@@ -47,6 +53,53 @@ export function isTypeForMinMax(type) {
...
+    allTabularDataTypes
+  )
+
+  const totalTypes = documentTypes.concat(imageTypes).concat(tabularDataTypes)
+
+  if (totalTypes.length) {
+    return totalTypes.length > 1
+      ? totalTypes.slice(0, totalTypes.length - 1).join(', ') +
+          ` or ${totalTypes[totalTypes.length - 1]}`
+      : (totalTypes[0] ?? '')
```
whitewaterdesign, line 97: I think there's a cleaner way to do this - using a loop or map then just changing the second last item.

src/server/plugins/engine/components/MultilineTextField.ts
```diff
@@ -73,6 +73,13 @@ export class MultilineTextField extends FormComponent {
       })
     } else if (options.customValidationMessages) {
       formSchema = formSchema.messages(options.customValidationMessages)
+    } else if (
+      typeof schema.max === 'number' &&
+      typeof schema.min === 'number'
+    ) {
+      const minMaxErrorText = this.buildMinMaxText(schema.min, schema.max)
+      formSchema = formSchema.min(schema.min).message(minMaxErrorText)
```
davidjamesstone, line 81: Joi's ruleset could be used here: https://joi.dev/api/?v=17.13.3#anyruleset---aliases-

src/server/plugins/engine/components/FileUploadField.ts
```diff
@@ -272,7 +272,20 @@ export class FileUploadField extends FormComponent {
...
+        },
+        {
+          type: 'filesSize',
+          template: 'The selected file must be smaller than 100MB'
+        },
+        { type: 'filesEmpty', template: 'The selected file is empty' },
+        {
+          type: 'filesError',
+          template: 'The selected file could not be uploaded – try again'
+        }
```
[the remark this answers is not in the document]
jbarnsley10: Good spot. I'd missed the 'virus' error but I don;t have a listing of an error similar to 'The selected file has not fully uploaded'. Is that a legitimate error message?

src/server/plugins/error-preview/error-preview-helper.js
jbarnsley10: Nice. Will take a look

src/server/plugins/engine/components/MultilineTextField.ts
jbarnsley10: Nice. Will take a look

src/server/plugins/engine/components/FileUploadField.ts
jbarnsley10: It is a bit confusing but all the other error messages use 'limit' as their token, so without changing some common processing, it wouldn't be that easy to change the token name. It could be done though if you think it's too confusing.

src/server/plugins/engine/components/FileUploadField.ts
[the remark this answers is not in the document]
davidjamesstone: Yes it comes from [here](https://github.com/DEFRA/forms-runner/blob/main/src/server/plugins/engine/pageControllers/FileUploadPageController.ts#L48)

src/server/plugins/engine/components/MultilineTextField.ts
davidjamesstone: It's fine as-is BTW but could be helpful - see what you think.

jbarnsley10: Joi ruleset seems to work well. Wasn't aware it existsed

davidjamesstone (approved):

src/server/plugins/error-preview/error-preview-helper.js
whitewaterdesign: `const [last, penultimate, ...rest] = totalTypes.reverse()
const newLast = `${penultimate} or ${last}`
const arr = [...rest.reverse(), newLast]
return arr.join(',')`

whitewaterdesign: or even better...

```
const lastItem = totalTypes.pop()
const penultimate = totalTypes.pop()
return [...totalTypes, `${penultimate} or ${lastItem}`].join(',')
```

src/server/plugins/error-preview/error-preview-helper.js
```diff
@@ -90,17 +90,14 @@ export function lookupFileTypes(types) {
...
-      if (idx < totalTypes.length - 1) {
-        return `${acc}, ${el}`
-      }
-      return `${acc} or ${el}`
-    }
-    return el
-  }, '')
-
-  return finalList.length ? finalList : '[files types you accept]'
+  if (totalTypes.length) {
```
whitewaterdesign, line 93:
```
const totalTypes = documentTypes.concat(imageTypes).concat(tabularDataTypes)

  const lastItem = totalTypes.pop()

  if (!lastItem) {
    return '[files types you accept]'
  }

  const penultimate = totalTypes.pop()

  if (!penultimate) {
    return lastItem
  }

  return [...totalTypes, `${penultimate} or ${lastItem}`].join(',')
```

whitewaterdesign: how about that - very clear and very performant

PR #779, opened 2025-04-16
Fix/526386  add short desc summary

whitewaterdesign:
## Add short description to summary text

Missing AC from previous short description ticket.  

https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/526386


jbarnsley10 (approved):
Looks good. I'd prefer the 'short description' to be a bit more of an obvious difference to the question label/title so it's clear how short description is used, but I can't think of how to improve int his test example other than perhaps to call the short description 'Delivery or collection'. But not worth changing the whole test example to a different scenario just for that.

PR #782, opened 2025-04-17
Feature 533156: Add /form prefix to routes & enforce strict path validation

mokhld:
This PR prefixes all form routes with `/form` and implements strict path validation with legacy redirect support (which at some point could be removed 🤔 ?).

The legacy redirect logic resides within an `onRequest` hook that intercepts incoming `GET` requests:

1.  **Ignored Prefixes:** Paths already starting with `/form/`, `/help/`, `/assets/`, etc., along with the root path `/`, are immediately passed through (`h.continue`).
2.  `/preview/{state}/{slug}`: Paths matching this are validated so that `state` must be 'live' or 'draft', and the `slug` must be conformant to the `slugSchema` (lowercase letters, numbers, hyphens).
    *   Valid paths (e.g., `/preview/live/my-slug`) are permanently redirected to `/form/preview/live/my-slug`.
    *   Paths with an invalid state or slug (e.g., `/preview/invalid/my-slug` or `/preview/live/InvalidSlug`) result in a 404.
3.  `/{slug}/{path...}`: Paths matching this  have their initial `slug` segment validated against `slugSchema`.
    *   Valid paths (e.g., `/my-slug/page-one`) are permanently redirected to `/form/my-slug/page-one`.
    *   If the slug is invalid (e.g., `/Invalid-Slug/page`), the redirect is skipped, and the request continues (likely resulting in a 404 from the router).
4.  `/{slug}`: Paths containing only a valid `slug` (matching regex and `slugSchema`, e.g., `/my-slug`) are permanently redirected to `/form/my-slug`.
5.  Other Paths: Any remaining paths that don't match the ignore list or the legacy redirect patterns (e.g., `/robots.txt`) fall through to `h.continue` and are handled by the router will typically resulting in a 404 (the original point of this story).

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -101,7 +101,7 @@ export async function createServer(routeConfig?: RouteConfig) {
 
   server.registerService(CacheService)
 
-  await server.register(pluginEngine)
+  await server.register(...pluginEngine)
 
   await server.register(pluginRouter)
 
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -101,11 +101,7 @@ export async function createServer(routeConfig?: RouteConfig) {
 
   server.registerService(CacheService)
 
-  await server.register(pluginEngine, {
-    routes: {
-      prefix: '/form'
-    }
-  })
+  await server.register(pluginEngine)
 
   await server.register(pluginRouter)
 
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -1,3 +1,5 @@
+import { slugSchema } from '@defra/forms-model'
+import Boom from '@hapi/boom'
 import { Engine as CatboxMemory } from '@hapi/catbox-memory'
 import { Engine as CatboxRedis } from '@hapi/catbox-redis'
 import hapi, {
@@ -24,6 +26,7 @@ import { plugin as pluginViews } from '~/src/server/plugins/nunjucks/index.js'
 import pluginPulse from '~/src/server/plugins/pulse.js'
 import pluginRouter from '~/src/server/plugins/router.js'
 import pluginSession from '~/src/server/plugins/session.js'
+import { stateSchema } from '~/src/server/schemas/index.js'
 import { prepareSecureContext } from '~/src/server/secure-context.js'
 import { CacheService } from '~/src/server/services/index.js'
 import { type RouteConfig } from '~/src/server/types.js'
@@ -98,6 +101,14 @@ export async function createServer(routeConfig?: RouteConfig) {
 
   server.registerService(CacheService)
 
+  await server.register(pluginEngine, {
+    routes: {
+      prefix: '/form'
+    }
+  })
+
+  await server.register(pluginRouter)
+
   server.ext('onPreResponse', (request: Request, h: ResponseToolkit) => {
     const { response } = request
 
@@ -121,8 +132,6 @@ export async function createServer(routeConfig?: RouteConfig) {
   })
 
   await server.register(pluginViews)
-  await server.register(pluginEngine)
-  await server.register(pluginRouter)
   await server.register(pluginErrorPages)
   await server.register(blipp)
   await server.register(requestTracing)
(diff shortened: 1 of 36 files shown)


src/server/plugins/engine/helpers.ts
```diff
@@ -254,22 +254,51 @@ export function getStartPath(model?: FormModel) {
...
+  const formRoutePrefix = '/form'
+
+  let isPreview = false
+  let stateSegmentIndex = -1
+  const lowerCasePath = path.toLowerCase()
+
+  if (lowerCasePath.startsWith(formRoutePrefix + previewPrefix)) {
+    isPreview = true
+    // State is after ['', 'form', 'preview', 'STATE']
+    stateSegmentIndex = 3
```
alexluckett, line 267: what if the prefix is `/many/nested/levels`?

mokhld: Hmm that's a _very_ good point

mokhld: We could actually revert this as the `onRequest` hook handles redirecting legacy `/preview/...` URLs first, so `checkFormStatus` only needs to parse the final `/form/preview/...` path if that makes sense.

src/server/index.ts
```diff
@@ -136,5 +141,83 @@ export async function createServer(routeConfig?: RouteConfig) {
     encoding: 'none' // handle this inside the application so we can share frontend/backend cookie modification
   })
 
+  server.ext('onRequest', (request: Request, h: ResponseToolkit) => {
+    const { method, path } = request
```
alexluckett, line 145: Feels a bit heavy handed to do this on every request. We could set up a route hander to mimic the old live forms, which would avoid us needing to ignore a set of prefixes?

```
{
  method: 'GET',
  path: '/{slug}',
  handler: (request) => {
    if (request.params.slug is a valid slug):
       return h.redirect(`forms/${slug}`)
    } else {
       return boom.notfound()
   }
  }
}
```

src/server/plugins/router.ts
```diff
@@ -22,14 +31,72 @@ import {
...
     name: 'router',
     register: (server) => {
       server.route(routes)
 
+      // /preview/{state}/{slug} -> {FORM_PREFIX}/preview/{state}/{slug}
+      server.route({
+        method: 'GET',
+        path: '/preview/{state}/{slug}',
+        options: {
+          handler: (request: Request, h: ResponseToolkit) => {
```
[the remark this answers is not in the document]
mokhld: Good spot @davidjamesstone , that's been shifted up.

src/server/plugins/engine/plugin.ts
```diff
@@ -129,9 +130,11 @@ export const plugin = {
...
 
         // Set up the basePath for the model
-        const basePath = isPreview
-          ? `${PREVIEW_PATH_PREFIX.substring(1)}/${formState}/${slug}`
-          : slug
+        const basePath = (
+          isPreview
+            ? `${prefix}${PREVIEW_PATH_PREFIX}/${formState}/${slug}`
+            : `${prefix}/${slug}`
+        ).substring(1)
```
alexluckett, line 137: What if the prefix doesn't begin with `/`? Seems a bit dangerous, unless Hapi itself forces the `/` prefix?

alexluckett: If hapi accepts `forms` as a prefix, I'd suggest we either throw an error when the plugin is initialised or wrap this logic in a function `safeReplacePrefix` that checks before dropping it.

davidjamesstone: Hapi enforces a leading forward slash [through Joi validation](https://github.com/hapijs/hapi/blob/76e6cfeb4d6a4f7d5b945f9c50ac8e12b84e32a5/lib/config.js#L400)

alexluckett: Perfect 👌 thanks Dave

alexluckett (approved):

davidjamesstone:
@alexluckett you're right. @mokhld I think all that's needed is to use the `checkFormStatus` function [here](https://github.com/DEFRA/forms-runner/blob/feature/strict-path-validation/src/server/plugins/nunjucks/context.js#L40), instead of using `const isPreviewMode = path?.startsWith(PREVIEW_PATH_PREFIX)`

alexluckett (approved):

PR #785, opened 2025-04-24
Feature/strict path validation 2

davidjamesstone:
@mokhld I've been taking a look at your PR and have been testing out a couple of commits here.

The first [use prefix from server.realm](https://github.com/DEFRA/forms-runner/pull/785/commits/9b21c9c8a0cf8b210bcce65aa7121c0ac4254175) - rather than use the constant prefix "form" inside the plugin, we can use hapi's `server.realm.modifiers.route.prefix`, that way it'll work if the plugin is registered with something other than "/form" as the prefix.

The second commit [simplify the checkFromStatus method](https://github.com/DEFRA/forms-runner/pull/785/commits/361b505c3733d0c84a078c0ebcf7b41115b1adcf) is something I've been meaning to look at for a while as it can be much simpler. We know we're in "preview mode" simply by looking if there's a `state` entry in the params.

Speak tomorrow :wave:

### DIFF
--- a/src/server/plugins/engine/helpers.test.ts
+++ b/src/server/plugins/engine/helpers.test.ts
@@ -3,7 +3,6 @@ import { type ResponseObject, type ResponseToolkit } from '@hapi/hapi'
 import { StatusCodes } from 'http-status-codes'
 import { ValidationError } from 'joi'
 
-import { PREVIEW_PATH_PREFIX } from '~/src/server/constants.js'
 import {
   checkEmailAddressForLiveFormSubmission,
   checkFormStatus,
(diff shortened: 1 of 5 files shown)


mokhld:
>The first [use prefix from server.realm](https://github.com/DEFRA/forms-runner/pull/785/commits/9b21c9c8a0cf8b210bcce65aa7121c0ac4254175) - rather than use the constant prefix "form" inside the plugin, we can use hapi's server.realm.modifiers.route.prefix, that way it'll work if the plugin is registered with something other than "/form" as the prefix.

This makes sense: leveraging Hapi's server.realm and removing the redundant constant 👍 

>The second commit [simplify the checkFromStatus method](https://github.com/DEFRA/forms-runner/pull/785/commits/361b505c3733d0c84a078c0ebcf7b41115b1adcf) is something I've been meaning to look at for a while as it can be much simpler. We know we're in "preview mode" simply by looking if there's a state entry in the params.

This is much cleaner and I suppose shifts the dependency on the route definition.

mokhld:
Cherry-picked these commits in https://github.com/DEFRA/forms-runner/pull/782

PR #802, opened 2025-06-02
Feat/562471 list id linkage

jbarnsley10:
Component to List linkage is now via 'id' for V2 schemas (was previously done using name')
There is a lint warning in FormsModel.ts but another PR is incoming from another dev to fix that

### DIFF
--- a/src/server/plugins/engine/models/FormModel.ts
+++ b/src/server/plugins/engine/models/FormModel.ts
@@ -4,6 +4,7 @@ import {
   ControllerPath,
   ControllerType,
   Engine,
+  SchemaVersion,
   formDefinitionSchema,
   hasComponents,
   hasRepeater,
@@ -56,6 +57,7 @@ export class FormModel {
   /** the entire form JSON as an object */
   def: FormDefinition
 
+  schemaVersion?: SchemaVersion
   lists: FormDefinition['lists']
   sections: FormDefinition['sections'] = []
   name: string
@@ -109,6 +111,7 @@ export class FormModel {
     setPageTitles(def)
 
     this.engine = def.engine
+    this.schemaVersion = def.schema
     this.def = def
     this.lists = def.lists
     this.sections = def.sections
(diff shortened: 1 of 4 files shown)


src/server/plugins/engine/models/FormModel.ts
```diff
@@ -281,8 +282,10 @@ export class FormModel {
     return parser.parse(conditions.toExpression())
   }
 
-  getList(name: string): List | undefined {
-    return this.lists.find((list) => list.name === name)
+  getList(nameOrId: string): List | undefined {
+    return this.schemaVersion === SchemaVersion.V1 && nameOrId !== yesNoListId
```
alexluckett, line 286: we can drop this check on yesnolist and just keep it to "if schema v1 get by name, if schema v2 get by id"

src/server/plugins/engine/components/YesNoField.ts
```diff
@@ -17,7 +19,7 @@ export class YesNoField extends SelectionControlField {
     def: YesNoFieldComponent,
     props: ConstructorParameters<typeof SelectionControlField>[1]
   ) {
-    super({ ...def, list: '__yesNo' }, props)
+    super({ ...def, list: yesNoListId }, props)
```
alexluckett, line 22: with my below comment, we could do:

```javascript
export const yesNoListId = '3167ecb5-61f9-4918-b7d0-6793b56aa814'
export const yesNoListName = '__yesNo'
```

```javascript
super({ ...def, list: def.schema === 1 : yesNoName ? yesNoListId }, props)
```

src/server/plugins/engine/models/FormModel.ts
```diff
@@ -108,7 +109,7 @@ export class FormModel {
 
     // Add default lists
     def.lists.push({
-      id: '3167ecb5-61f9-4918-b7d0-6793b56aa814',
+      id: yesNoListId,
```
alexluckett, line 112:
```javascript
id: def.schema === 1 : yesNoName ? yesNoListId,
````

src/server/plugins/engine/models/FormModel.ts
alexluckett: the reason I suggest this, is we don't know what other lists might be pushed in the future - a consumer "getting a list" shouldn't really need to worry about that kind of lower level detail. It's really a responsibility of the unit pushing in the list to conform, not the getter.

PR #804, opened 2025-06-06
Feat/551140  upgrade to govuk fe 5

whitewaterdesign:
## We need to update to the new govuk branding on 25th June

- environment variable for the update - `SHOW_GOVUK_REBRAND`
- full width bottom bar on current version so we can include `govukServiceNavigation` now
- implement `govukServiceNavigation` now

Ticket: https://dev.azure.com/defragovuk/DEFRA-CDP/_workitems/edit/551140


jbarnsley10 (approved):
Looks good to me. I'd perhaps question the need for the 'rebrand' config flag though (unless that was part of the requirements). If we upgrade the npm, then the codebase uses the new branding (i.e. we don't have a choice). However, can be persuaded either way.

whitewaterdesign:
Note, if you're reviewing locally, there may be a caching issue with the gov uk tag.

whitewaterdesign:
> Looks good to me. I'd perhaps question the need for the 'rebrand' config flag though (unless that was part of the requirements). If we upgrade the npm, then the codebase uses the new branding (i.e. we don't have a choice). However, can be persuaded either way.

this is part of requirements

src/server/views/layout.html
```diff
@@ -118,16 +119,18 @@
       text: config.serviceBannerText
     }) }}
   {% endif %}
-
   {{ govukHeader({
     homepageUrl: currentPath if context.isForceAccess else "https://www.gov.uk",
     containerClasses: "govuk-width-container",
-    productName: productName | safe | trim,
+    classes: "govuk-header--full-width-border",
+    rebrand: govukRebrand,
```
alexluckett, line 126: is this required if the nunjucks global is set?

alexluckett (approved):

PR #806, opened 2025-06-12
Update toConditionContext ready for join conditions (ConditionRefDataV2)

davidjamesstone:
Expose conditions as getters using the condition displayName rather than the key (uuid) in condition context


alexluckett (dismissed):

davidjamesstone:
Converted back to draft PR - we will pick this up when we do the "join conditions" feature.

It may be better to give each condition an random alias that is applied to the condition context as a getter - this way we'll not run into issues with using the name or uuid that could break expr-eval when it encounters a condition ref e.g. `'(isBobV2 and (isFaveColourRedV2))'`

PR #813, opened 2025-06-19
Fix/571592 date conditions

jbarnsley10:
This was a bit more complicated than first thought.

The time element has to be stripped (as expected) but then toISOString() was potentially causing the wrong day to be selected due to BST e.g. 2025-06-19T23:00:00.000Z.
Therefore, after stripping the time, the date is formatted as YYYY-MM-DD to ensure the exact correct date/time is used for the comparison.

There is an override function to allow proper unit testing of this functionality. Alex's example form definition is used as the basis for a series of relative date tests (overriding the current time to be fixed for the tests).

### DIFF
--- a/src/server/plugins/engine/models/FormModel.ts
+++ b/src/server/plugins/engine/models/FormModel.ts
@@ -20,7 +20,7 @@ import {
   type List,
   type Page
 } from '@defra/forms-model'
-import { add, format, startOfToday } from 'date-fns'
+import { add, format } from 'date-fns'
 import { Parser, type Value } from 'expr-eval'
 import joi from 'joi'
 
@@ -33,6 +33,7 @@ import {
   hasListFormField,
   type Component
 } from '~/src/server/plugins/engine/components/helpers.js'
+import { todayAsDateOnly } from '~/src/server/plugins/engine/date-helper.js'
 import {
   findPage,
   getError,
@@ -229,10 +230,7 @@ export class FormModel {
    * Instantiates a Condition based on {@link ConditionWrapper}
    * @param condition
    */
-  makeCondition(
-    condition: ConditionWrapper,
-    currentTimeFn?: () => Date // For unit testing
-  ): ExecutableCondition {
+  makeCondition(condition: ConditionWrapper): ExecutableCondition {
     const parser = new Parser({
       operators: {
         logical: true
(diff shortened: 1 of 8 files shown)


src/server/plugins/engine/models/FormModel.ts
```diff
@@ -229,7 +229,10 @@ export class FormModel {
    * Instantiates a Condition based on {@link ConditionWrapper}
    * @param condition
    */
-  makeCondition(condition: ConditionWrapper): ExecutableCondition {
+  makeCondition(
+    condition: ConditionWrapper,
+    currentTimeFn?: () => Date // For unit testing
+  ): ExecutableCondition {
```
alexluckett, line 235:
```suggestion
  makeCondition(
    condition: ConditionWrapper,
    currentTimeFn: () => Date = startOfToday // For unit testing
  ): ExecutableCondition {
```

src/server/plugins/engine/models/FormModel.ts
```diff
@@ -238,7 +241,16 @@ export class FormModel {
...
+        // The time element must be stripped (hence using startOfDay() which has no time element),
+        // then formatted as YYYY-MM-DD otherwise we can hit time element and BST issues giving the
+        // wrong date to compare against.
+        // Do not use .toISOString() to format the date as that introduces BST errors.
+        return format(
+          add(currentTimeFn !== undefined ? currentTimeFn() : startOfToday(), {
+            [timeUnit]: timePeriod
+          }),
+          'yyyy-MM-dd'
+        )
```
alexluckett, line 253:
```suggestion
        return format(
          add(currentTimeFn(), {
            [timeUnit]: timePeriod
          }),
          'yyyy-MM-dd'
        )
```

src/server/plugins/engine/models/FormModel.ts
alexluckett: Bit cleaner to use default args like this + the below comment?

alexluckett: Although, can't we just mock `startOfToday` using jest rather than needing to swap in the function?

jbarnsley10: Yep, good thinking. Originally it was 'new Date()' with a stripTimeFromDate() method (hence not being able to mock), but now using startOfToday (Dave's suggestion) so I can mock it now.

alexluckett (approved):

PR #893, opened 2025-08-29
Feature/DF- 370: save form progress 2

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-370


src/server/plugins/router.ts
```diff
@@ -398,13 +398,13 @@ export default {
           // Get the flashed email
           const messages = request.yar.flash(getFlashKey(params))
```
alexluckett, line 399: Out of scope of this PR but using form ID would be better, in case the form gets renamed?

alexluckett (approved):
LGTM

PR #903, opened 2025-09-03
Feature/DF-370 save form progress 4

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-370


src/server/routes/save-and-exit.js
```diff
@@ -0,0 +1,290 @@
...
+        return h
+          .redirect(`${ERROR_BASE_URL}/${form.slug}`)
+          .code(StatusCodes.SEE_OTHER)
+      }
+
+      const { isPreview, status } = linkDetails.form
+
+      const slugAndState = isPreview ? `/${status}` : ''
+
+      return h.redirect(
```
alexluckett, line 93: What about if the form ID in the path params doesn't match the form ID from linkDetails?

src/server/messaging/publish.test.js
```diff
@@ -13,17 +13,22 @@ import { publishSaveAndExitEvent } from '~/src/server/messaging/publish.js'
 
 jest.mock('~/src/server/messaging/publish-base.js')
 
+/**
+ * @type {import('@defra/forms-model').SaveAndExitMessageData}
```
whitewaterdesign, line 17: should this be at the bottom?

alexluckett:
I've resolved the Sonar security false-positive. It'll disappear on the next scan.

Looks like you're missing a test to cover `/resume-form-verify`, which would boost the coverage quite a bit.

whitewaterdesign (approved):
👍

PR #911, opened 2025-09-10
Fix/df 372 resume save and exit page titles

davidjamesstone:
https://eaflood.atlassian.net/browse/DF-372

Add missing page title and relocate S&E view files into the S&E directory


jbarnsley10 (approved):
Looks good. Nice that we have all the html S&E pages together now

PR #1146, opened 2026-04-01
feat: add environment configuration changes

parveshh:


src/config/index.ts
```diff
@@ -89,30 +98,29 @@ export const config = convict({
...
 
   /**
    * Session storage
    * Redis integration is optional, but recommended for production environments.
    */
   sessionTimeout: {
     format: Number,
-    default: oneHour * 24, // 1 day
+    default: oneHour * 24,
     env: 'SESSION_TIMEOUT'
```
alexluckett, line 114: why was this not moved too?

src/config/index.ts
```diff
@@ -50,14 +50,23 @@ export const config = convict({
...
+    | 'dev'
+    | 'test'
+    | 'perf-test'
+    | 'ext-test'
+    | 'prod'
+  >,
   enforceCsrf: {
     format: Boolean,
     default: true,
     env: 'ENFORCE_CSRF'
```
alexluckett, line 65: why was this not moved too?

src/config/index.ts
```diff
@@ -89,30 +98,29 @@ export const config = convict({
...
     format: Number,
-    default: oneHour * 24, // 1 day
+    default: oneHour * 24,
     env: 'SESSION_TIMEOUT'
-  },
+  } as SchemaObj<number>,
   confirmationSessionTimeout: {
     format: Number,
     default: oneMinute * 20,
     env: 'CONFIRMATION_SESSION_TIMEOUT'
```
alexluckett, line 119: why was this not moved too?

src/config/index.ts
```diff
@@ -263,94 +271,90 @@ export const config = convict({
 
   safelist: {
     format: Array,
-    default: ['61bca17e-fe74-40e0-9c15-a901ad120eca.mock.pstmn.io'],
+    default: null,
     env: 'SAFELIST'
-  },
```
alexluckett, line 268: I think this is an unused config. It can probably be removed - check though.

src/config/index.ts
```diff
@@ -263,94 +271,90 @@ export const config = convict({
...
+    default: null,
     env: 'GOOGLE_ANALYTICS_CONTAINER_ID'
-  },
+  } as SchemaObj<string>,
 
   saveAndExitExpiryDays: {
     format: Number,
-    default: 28,
+    default: null,
     env: 'SAVE_AND_EXIT_EXPIRY_IN_DAYS'
```
alexluckett, line 300: this is missing in cdp-app-config's defaults

alexluckett (changes requested):

src/config/index.ts
parveshh: could not any use of this in codebase so removed

alexluckett (approved):

PR #1196, opened 2026-05-13
feat(DF-789): block save-and-exit and help routes when form is offline

mokhld:
Service-layer wrappers in `formMetadataGuards.js` throw the offline marker; engine plugin's `onPreResponse` renders the unavailable view.
References offline guards from engin-plugin

### DIFF
--- a/src/server/services/formMetadataGuards.js
+++ b/src/server/services/formMetadataGuards.js
@@ -1,0 +1,60 @@
+import Boom from '@hapi/boom'
+
+import * as rawFormsService from '~/src/server/services/formsService.js'
+
+/**
+ * Throws a Boom 503 stamped with `data.offline = true` when the form is
+ * offline. The engine plugin's onPreResponse extension catches the marker
+ * and renders the unavailable view at HTTP 200.
+ * @param {FormMetadata} metadata
+ */
+function assertFormAvailable(metadata) {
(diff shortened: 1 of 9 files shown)


src/server/services/formMetadataGuards.js
```diff
@@ -0,0 +1,60 @@
...
+
+import * as rawFormsService from '~/src/server/services/formsService.js'
+
+/**
+ * Throws a Boom 503 stamped with `data.offline = true` when the form is
+ * offline. The engine plugin's onPreResponse extension catches the marker
+ * and renders the unavailable view at HTTP 200.
+ * @param {FormMetadata} metadata
+ */
+function assertFormAvailable(metadata) {
```
jbarnsley10: Can these two guard methods be pulled from plugin, so they don't have to be duplicated here? Not a biggy though

jbarnsley10 (approved):
Looks good. A v minor comment but approving in principal

src/server/services/formMetadataGuards.js
mokhld: Possibly, but it will be a bit more of a refactor. I don't think I have time for it. Possibly tech debt that we can come back to?

jbarnsley10: Now references them from plugin

PR #1241, opened 2026-06-26
Model and Plugin Updates (& fix test)

Scullyon:
Updating model and runner so that we can include a range of conditions targeting a list question


davidjamesstone (approved):
Same as the other PR - version shouldn't contain a `v`

PR #1242, opened 2026-06-30
DF-1069: feature/i18n-welsh

davidjamesstone:
Built on https://github.com/DEFRA/forms-engine-plugin/pull/436 (v5.0.0-alpha)

runner translation implementation for layouts and routes not provided by engine-plugin


src/server/plugins/router.ts
```diff
@@ -123,8 +124,13 @@ export default {
         async handler(request, h) {
           const { slug } = request.params
           const form = await getFormMetadataWithoutGuard(slug)
+          const { translator } = await getFormTranslator(
+            request,
+            form,
+            FormStatus.Draft
```
davidjamesstone: Hardcoded to `FormStatus.Draft` - is that correct or could we do what we do in other places in this file: `const formStatus = form.live ? FormStatus.Live : FormStatus.Draft`?

jbarnsley10: Yes, I think you're right. There are about 7 places where that should probably be changed

jbarnsley10: It's only for the cache key, but should probably be consistent

src/server/routes/save-and-exit.js
```diff
@@ -244,7 +299,7 @@ export default [
         return Boom.badRequest('No email found in session cache')
       }
 
-      const model = confirmationViewModel(metadata, email, status)
+      const model = confirmationViewModel(metadata, email, translator, status)
```
alexluckett, line 302: `language` is missing from here so it always uses english

However I'm not sure a `language = 'en-GB'` default is sensible anyway, so I think there needs to be another fix here: https://github.com/DEFRA/forms-runner/pull/1242#discussion_r3614388340

src/server/models/save-and-exit.js
```diff
@@ -368,18 +420,27 @@ export function detailsViewModel(metadata, status, payload, err) {
...
  * @param {FormStatus} [status]
+ * @param {string} [language]
  */
-export function confirmationViewModel(metadata, email, status) {
+export function confirmationViewModel(
+  metadata,
+  email,
+  translator,
+  status,
+  language = 'en-GB'
```
alexluckett: `language` is never passed in, so this always defaults to English.

I'd avoid using `language` as a function-level fallback. Use a single source of truth.

src/server/models/save-and-exit.js
```diff
@@ -30,28 +27,64 @@ const GOVUK_LABEL__M = 'govuk-label--m'
...
+  return /** @type {string} */ (t(detail.message, language))
+}
 
 /**
  * Build form errors
  * @param {Error} [err]
+ * @param {string} [language]
  */
-function buildErrors(err) {
+function buildErrors(err, language = 'en-GB') {
```
alexluckett: language fallback required when it's already built into the translator?

src/server/models/save-and-exit.js
alexluckett: is the language fallback required when it's already built into the translator?

src/server/models/save-and-exit.js
```diff
@@ -389,125 +450,152 @@ export function confirmationViewModel(metadata, email, status) {
...
 
 /**
  * The save and exit error form view model
  * @param {{ slug: string }} payload
+ * @param {Translator} [translator] - the translator instance
  */
-export function resumeErrorViewModel(payload) {
-  const pageTitle = 'You cannot resume your form'
+export function resumeErrorViewModel(payload, translator) {
+  const { language = 'en-GB' } = translator ?? {}
```
alexluckett: language fallback required when it's already built into the translator?

alexluckett (commented):
to clarify, what i mean is that `createTranslator` in `src/server/i18n/form.ts` already embeds the language and returns a closure `t` which doesn't require the language to be passed in at every call site, so all the `language = 'en-GB'` fallbacks aren't required because `t` would already have it embedded.

jbarnsley10 (approved):

PR #1248, opened 2026-07-23
feat/df-1063: Passes selected language to downstream systems

jbarnsley10:
Passes selected language to downstream systems
Translates the Alpha/Beta banner if necessary

Needs a plugin bump when https://github.com/DEFRA/forms-engine-plugin/pull/445 is merged


src/server/plugins/SummaryPageWithConfirmationEmailController.test.ts
```diff
@@ -61,7 +62,7 @@ describe('SummaryPageWithConfirmationEmailController', () => {
       },
       query: {},
       app: { model }
-    } as unknown as FormRequest)
```
alexluckett, line 64: 👌

davidjamesstone (approved):

src/server/services/outputService.ts
```diff
@@ -42,6 +42,7 @@ export class OutputService implements IOutputService {
     _emailAddress: string,
     items: DetailItem[],
     submitResponse: SubmitResponsePayload,
+    _language: string,
```
alexluckett: rather than adding this as a parameter, you could make use of the request.server property? `getPluginOptions(request.server).getLanguage(request)` 

Would have saved your sonar issue 😀

davidjamesstone (approved):

PR #1250, opened 2026-07-28
fix/df-1069: Language selection fix when in mini-journey

jbarnsley10:
Fix to postcode lookup translations
- allows switch of language when in a mini journey

### DIFF
--- a/src/server/plugins/nunjucks/environment.js
+++ b/src/server/plugins/nunjucks/environment.js
@@ -31,5 +31,6 @@ environment.addFilter('evaluate', (value) => value)
 environment.addFilter('markdown', (value, startingHeaderLevel = 1) =>
   markdownToHtml(value, { startingHeaderLevel })
 )
+environment.addFilter('applyUrlParam', (value) => value)
 
 environment.addGlobal('govukRebrand', true)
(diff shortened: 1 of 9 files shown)


davidjamesstone (approved):

src/server/plugins/nunjucks/environment.js
```diff
@@ -31,5 +31,6 @@ environment.addFilter('evaluate', (value) => value)
 environment.addFilter('markdown', (value, startingHeaderLevel = 1) =>
   markdownToHtml(value, { startingHeaderLevel })
 )
+environment.addFilter('applyUrlParam', (value) => value)
```
alexluckett, line 34: can be dropped as it doesn't do anything?

alexluckett: or is applyUrlParam a plugin function and you're patching on non-plugin pages?

jbarnsley10: `applyUrlParam` is used in the language-selector (but the actual filter is defined in the `plugin`)

davidjamesstone: I asked the same, although I'm not sure where that comment has gone. Jez said it's used in runner.

PR #1252, opened 2026-07-29
feat/df-1069: Defaults to English if no translations

jbarnsley10:
Defaults to English if no translations configured for the form

### DIFF
--- a/src/server/routes/save-and-exit.js
+++ b/src/server/routes/save-and-exit.js
@@ -47,7 +47,10 @@ import {
   validateSaveAndExitCredentials
 } from '~/src/server/services/formsService.js'
 import { getFormDefinitionWithFallback } from '~/src/server/services/helpers/formsServiceHelper.js'
-import { resolveLanguage } from '~/src/server/utils/utils.js'
+import {
+  isLanguageSupported,
+  resolveLanguage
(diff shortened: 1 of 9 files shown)


src/server/routes/save-and-exit.js
```diff
@@ -90,7 +95,16 @@ export async function getFormTranslator(
...
+  let language = resolveLanguage(request.query, request.yar)
+
+  if (language !== EN_GB) {
+    const definition = await getFormDefinitionWithFallback(metadata.id, status)
+
+    if (!isLanguageSupported(language, definition)) {
+      // If not translations defined in the FormDefinition, always default to English
+      language = EN_GB
+    }
+  }
```
alexluckett, line 107: do we need this defined here if we already have it handled in the nunjucks context?

alexluckett (approved):

src/server/routes/save-and-exit.js
jbarnsley10: Good suggestion, but when running save-and-exit routes, the `context` doesn't contain the formId or definition, so we have to hit an additional method that can get the form def (async) and check the supported languages

alexluckett: ah of course

PR #1257, opened 2026-08-06
DF-1248: Add error details when previewing a form

alexluckett:
This PR depends on https://github.com/DEFRA/forms-engine-plugin/pull/459.

When a form is broken, the error page currently displays a standard HTTP 500 page. This PR introduces an error page specifically for form previews to tell the form designer what is broken.

This page is only shown when accessing a preview form. Live submissions will not show it.

This page is only displayed if the error is related to the form definition or form metadata. Standard errors during a preview will show up as the standard HTTP 500 page.

Jira ticket: [DF-1248](https://eaflood.atlassian.net/browse/DF-1248)

## Invalid form definition

image

## Invalid metadata

image

## Unknown page controllers

image

## Unknown components

image

[DF-1248]: https://eaflood.atlassian.net/browse/DF-1248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "@hapi/wreck": "^18.1.2",
     "@hapi/yar": "^11.0.3",
     "@types/humanize-duration": "^3.27.4",
-    "accessible-autocomplete": "^3.0.1",
     "babel-plugin-transform-import-meta": "^2.3.3",
     "blankie": "^5.0.0",
     "blipp": "^4.0.2",
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
   "license": "SEE LICENSE IN LICENSE",
   "dependencies": {
     "@aws-sdk/client-sns": "^3.1079.0",
-    "@defra/forms-engine-plugin": "5.0.0-alpha.12",
-    "@defra/forms-model": "^3.0.693",
+    "@defra/forms-engine-plugin": "^5.0.0-alpha.13",
+    "@defra/forms-model": "^3.0.694",
     "@defra/hapi-tracing": "^1.30.0",
     "@elastic/ecs-pino-format": "^1.5.0",
     "@hapi/boom": "^10.0.1",
@@ -62,6 +62,7 @@
     "@hapi/wreck": "^18.1.2",
     "@hapi/yar": "^11.0.3",
     "@types/humanize-duration": "^3.27.4",
+    "accessible-autocomplete": "^3.0.1",
     "babel-plugin-transform-import-meta": "^2.3.3",
     "blankie": "^5.0.0",
     "blipp": "^4.0.2",
--- a/src/server/i18n/translations/en-GB.json
+++ b/src/server/i18n/translations/en-GB.json
@@ -241,7 +241,7 @@
       "tryBack": "try your browser back button as your data may be saved",
       "contact": "<a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/contact-defra\">contact the Defra Helpline</a> if you have any questions",
       "details": {
-        "summary": "What went wrong (preview only)",
+        "summary": "What went wrong",
         "intro": "This information is shown because you are previewing this form. Members of the public will not see it.",
         "technicalHeading": "Technical details"
       }
--- a/src/server/i18n/translations/en-GB.json
+++ b/src/server/i18n/translations/en-GB.json
@@ -242,9 +242,9 @@
       "contact": "<a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/contact-defra\">contact the Defra Helpline</a> if you have any questions"
     },
     "previewError": {
-      "heading": "This form cannot be previewed",
+      "heading": "This form can't be previewed",
       "summary": "What went wrong",
-      "intro": "The form contains configuration errors that need to be resolved before it can be viewed by the public.",
+      "intro": "Something in the form setup needs attention before you can preview it.",
       "technical": {
         "revealText": "Technical details",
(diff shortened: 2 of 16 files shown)


src/server/i18n/translations/en-GB.json
```diff
@@ -240,6 +240,15 @@
       "canYou": "You can:",
       "tryBack": "try your browser back button as your data may be saved",
       "contact": "<a class=\"govuk-link\" href=\"https://www.gov.uk/guidance/contact-defra\">contact the Defra Helpline</a> if you have any questions"
+    },
```
jbarnsley10, line 243: You seem to have omitted the translations for the `cy.json` file. Since the error pages now (always) have a language selector - irrespective of whether the form has translations or not - we will need the Welsh translations for this page.

src/server/plugins/errorInterpretation.js
```diff
@@ -0,0 +1,150 @@
...
+
+  while (cause instanceof Error && depth < MAX_CAUSE_DEPTH) {
+    parts.push(`Caused by: ${cause.message}`)
+    cause = cause.cause
+    depth++
+  }
+
+  const text = parts.join('\n')
+
+  return text.length > MAX_TECHNICAL_LENGTH
```
jbarnsley10, line 36: The `ellipsis` can be done purely using CSS in the UI but I guess physically truncating the payload here is the correct approach to prevent bleed.

src/server/plugins/errorInterpretation.js
```diff
@@ -0,0 +1,150 @@
...
+      text: `The condition "${conditionName}" isn't configured correctly. Open the condition and check that:`,
+      items: [
+        'the question it refers to still exists',
+        'the correct answer option is selected',
+        'the condition has been completed and saved'
+      ]
+    }
+  },
+  [UnknownPageControllerError.name]: () =>
+    'This form uses a page type this version of the service does not recognise.',
```
jbarnsley10, line 60: Is it worth adding the value of the invalid controller name?

src/server/plugins/errorInterpretation.js
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * One message per validation failure in the form definition. A Set removes
+ * repeats, because several schema rules can produce the same message.
+ * @param {import('joi').ValidationError} joiError
+ * @returns {string[]}
+ */
+function buildDefinitionCauses(joiError) {
+  return [
+    ...new Set(
```
jbarnsley10, line 75: Nice

src/server/plugins/errorInterpretation.js
```diff
@@ -0,0 +1,150 @@
...
+ * One cause builder per known error type, keyed by the error's class name.
+ * To support a new error type, add an entry here. Errors without an entry
+ * fall back to their own message.
+ * @type {Record<string, ((error: InvalidFormDefinitionError) => string | ErrorCause) | undefined>}
+ */
+const causeBuildersByErrorName = {
+  [ConditionBuildError.name]: (error) => {
+    const { conditionName } = /** @type {ConditionBuildError} */ (error)
+    return {
+      text: `The condition "${conditionName}" isn't configured correctly. Open the condition and check that:`,
```
jbarnsley10, line 51: I'm happy with this not being translatable (since it's probably primarily for form designers, and they are designing in English) but just pointing out that the error page is translatable yet some of the detail won't be translated.

package.json
```diff
@@ -62,6 +62,7 @@
     "@hapi/wreck": "^18.1.2",
     "@hapi/yar": "^11.0.3",
     "@types/humanize-duration": "^3.27.4",
+    "accessible-autocomplete": "^3.0.1",
```
jbarnsley10: We shouldn't need to include `accessible-autocomplete` here. It's only when using `npm link` that it disappears (so I manually copy from plugin node-modules to runner node_modules temporarily). When using a proper plugin npm version, `accessible-autocomplete` is pulled in from plugin, and therefore not needed in the runner `package.json`

src/server/i18n/translations/en-GB.json
alexluckett: I actually left them off intentionally since this is for internal workers only, the language toggle will display but it won't do anything and will fall back to English. We don't actually need Welsh here.

src/server/plugins/errorInterpretation.js
alexluckett: I originally had it but I removed it because it's too technical and doesn't mean anything to the user, I left it in the technical details section only.

src/server/plugins/errorInterpretation.js
alexluckett: I was cognisant of this when building it, but ultimately decided it was okay since we're never going to support Welsh for Designer. Since I'm reusing errors from forms-model anyway that are hardcoded in English, it wasn't worth plumbing in the support for Welsh here. If I built support in here, I'd then need to look at adding support for the model errors and it wasn't worth it.

src/server/i18n/translations/en-GB.json
jbarnsley10: Would it be sensible to hide the language selector then on the error details page? That would just be a case of removing 'languages' from the viewModel, I think

package.json
alexluckett: Oops, thanks. Accidental commit from development.

jbarnsley10 (approved):

PR #1262, opened 2026-08-17
DF-1161: Sign in to Defra Forms

alexluckett:
Adds citizen sign in: an OpenID Connect relying party against forms-identity-ui, a per-form homepage that requires it, and the signed-in citizen's email in the service header. All of it sits behind `USE_SIGN_IN_FEATURE`, which defaults to off.

# Architecture

When `USE_SIGN_IN_FEATURE` is on, `createServer` registers two plugins and the router adds two route groups. 

`oidc-client` reads the OIDC settings, imports the signing key and holds the provider configuration. `auth` registers the `citizen-session` scheme and sets it as the server-wide default in `try` mode, so an anonymous request carries on with no credentials and every existing route keeps the behaviour it had. Using the default strategy means the plugin inherits the security without us needing to add it to the plugin explicitly.

The provider takes one client authentication method, `private_key_jwt`. This service holds the private half of an ES256 pair as a single JWK and signs a short-lived assertion with it; the provider holds the public half, and the `kid` in the assertion header tells it which one to verify with, so the provider can hold both halves of a rotation while this service signs with one. The key is imported at boot rather than at the first sign in, and the required settings are checked at the same point, so a key that cannot sign or a setting left empty stops the service starting and names itself.

Sign in is authorization code with PKCE. `/auth/sign-in` generates the state, nonce and code verifier, puts them in the session with the return path, and sends the citizen to the provider. `/auth/callback` matches the state against the session before it consumes the code, exchanges the code, then asks the userinfo endpoint for the user's email address. 

For security purposes to avoid malicious URL building, return targets go through a new `localReturnPath` function, which ensures that URLs only resolve to forms-runner. `returnUrlSchema` wraps this for route validation, so any malicious URLs are rejected with a HTTP 400 before the handler runs. The cookie-preferences route has been updated to use this new schema to avoid security issues on that endpoint too.

The service header is the One Login component with an adapted Nunjucks template. The published template names its account link "GOV.UK One Login" and points it at `home.account.gov.uk`, and exposes neither as an option, so the template is copied into `views/components/service-header/` and takes the citizen's email and their form homepage instead. The package stays a dependency for the stylesheet and the script that drives the dropdown at small screen sizes, even though the Nunjucks content has been duplicated.

# Key changes

- `USE_SIGN_IN_FEATURE` and four `OIDC_*` settings in config, with the environment variable table and a "Citizen sign in" section in the README, and entries in `.env.sample`.
- `plugins/oidc-client.ts` — provider configuration, client key, `checkConfigurationValid()` at registration.
- `plugins/auth.ts` and `auth/scheme.js` — the `citizen-session` scheme, its strategy, and the server-wide `try` default. The signed-in identity lands on `request.auth.credentials`.
- `auth/accountSession.js` — the reads and writes for the citizen identity and the in-flight sign in, with a documented reason for each field held.
- `routes/auth.js` — `/auth/sign-in` and `/auth/callback`.
- `routes/homepage.js` and `views/homepage.html` — `/homepage/{slug}`, which looks the form up first so an unknown slug answers 404 ahead of any trip through sign in.
- `utils/utils.js` — `localReturnPath` and `signInUrl`; `models/common.js` — `returnUrlSchema`, applied to `/auth/sign-in` and cookie preferences.
- `views/components/service-header/` and `_service-header.scss` — the header, rendered on every page, showing the account menu when the flag is on and a citizen is signed in.
- The Nunjucks context exposes `user`, `useSignInFeature` and `language`; new `signIn.*` strings in `en-GB.json`.


src/client/stylesheets/_service-header.scss
```diff
@@ -0,0 +1,15 @@
+// The One Login service header's Sass is written for the legacy `@import`
```
jbarnsley10: Sounds a bit AI-generated and some odd wording, but conveys the correct info, so fine as is

src/server/plugins/nunjucks/context.js
```diff
@@ -81,12 +83,16 @@ export function context(request) {
...
+      useSignInFeature: config.get('useSignInFeature'),
       feedbackViaEmail: config.get('feedbackViaEmail')
     },
     cspNonce: request?.plugins.blankie?.nonces?.script,
     currentPath: request ? `${request.path}${request.url.search}` : undefined,
     previewMode: isPreviewMode ? formState : undefined,
     slug: isResponseOK ? params?.slug : undefined,
+    user: credentials?.email ? { email: credentials.email } : null,
+
+    language,
```
jbarnsley10: We may want to add `languages[]` if the sign-in pages are to have a language selector - but for a later time when the translations are setup for Welsh

src/server/routes/auth.js
```diff
@@ -0,0 +1,171 @@
...
+      const nonce = client.randomNonce()
+
+      setSignInTransaction(request.yar, {
+        state,
+        nonce,
+        codeVerifier,
+        returnUrl
+      })
+
+      logger.info(
```
jbarnsley10: Do we need this logger entry? Will we be able to match a failure to a client's sessionId or state?

src/server/routes/auth.js
```diff
@@ -0,0 +1,171 @@
...
+        code_challenge_method: 'S256'
+      })
+
+      return h.redirect(authorizationUrl.href)
+    },
+    options: {
+      validate: {
+        // A citizen can arrive here from anywhere, and a link that picked up
+        // a tracking parameter on the way should still start a sign in. The
+        // return target is the only key this route reads, and it is validated
```
jbarnsley10: Is 'returnUrl' better than 'return target' ?

src/server/routes/auth.js
```diff
@@ -0,0 +1,171 @@
...
+    async handler(request, h) {
+      const transaction = getSignInTransaction(request.yar)
+
+      // `state` is the random value sign in generated and the provider sends
+      // back. It matches the one in this session when the callback belongs to
+      // this session's sign in. The transaction is read but not cleared, so a
+      // callback that does not match leaves it for the one that does.
+      if (!transaction || transaction.state !== request.query.state) {
+        logger.warn(
+          {
```
jbarnsley10: BTW I have a feeling these attributes won;t get persisted in the CDP logs. We have to save attributes under an `event` structure against specific sub-properties if we want to be able to see them in the logs e.g. what we did for payment in `plugin` https://github.com/DEFRA/forms-engine-plugin/blob/b815a0fea8a8080bdd51affdc5a907ebfc2e6d29/src/server/plugins/engine/routes/payment-helper.js#L50

src/server/routes/homepage.js
```diff
@@ -0,0 +1,41 @@
...
+      // 404 instead of a trip through sign in.
+      const form = await getFormMetadata(slug)
+
+      if (!request.auth.isAuthenticated) {
+        return h.redirect(signInUrl(`${HOMEPAGE_PREFIX}/${slug}`))
+      }
+
+      return h.view('homepage', {
+        form,
+        startUrl: `${FORM_PREFIX}/${slug}`
```
jbarnsley10: How are we handling draft/live forms, or are we only ever logging in for live forms?

src/server/views/components/service-header/template.njk
```diff
@@ -0,0 +1,133 @@
...
+{%- set signOutLink = params.signOutLink -%}
+
+{%- macro oneLoginIcon(modifier="default") -%}
+  {%- set class = "rebranded-cross-service-header__logo--nav" if modifier == "nav" else "" %}
+  <svg focusable="false" xmlns="http://www.w3.org/2000/svg" width="15" height="30" viewBox="150 150 250 250" fill="currentcolor" class="rebranded-cross-service-header__logo {{class}}" aria-hidden="true">
+    <path d="M300.002 312.261C249.445 312.261 208.346 271.165 208.346 220.608C208.346 170.051 249.445 128.954 300.002 128.954C350.559 128.954 391.655 170.051 391.655 220.608C391.655 271.165 350.559 312.261 300.002 312.261ZM300.002 170.892C272.673 170.892 250.389 193.175 250.389 220.504C250.389 247.83 272.673 270.113 300.002 270.113C327.33 270.113 349.611 247.83 349.611 220.504C349.611 193.175 327.33 170.892 300.002 170.892Z" />
+    <path d="M221.275 471.046H179.231V365.202H420.769V407.246H221.275V471.046Z"/>
+  </svg>
+{%- endmacro -%}
+<header class="rebranded-cross-service-header" data-module="one-login-header">
```
jbarnsley10: As long as we're not going to confuse people by having references to 'OneLogin' here. I guess it's covered by the comments at the top of the file anyway

src/server/views/homepage.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set pageTitle = form.title %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l">{{ tR('signIn.homepage.caption') }}</span>
+      <h1 class="govuk-heading-l">{{ form.title }}</h1>
```
jbarnsley10: This will eventually need to handle translation

jbarnsley10 (approved):
Excellent work - as always. Feels quite AI-written but probably mainly due to the comments. Code-wise it's very good. A possible gotcha about logging, and a few minor comments.

src/server/routes/homepage.js
alexluckett: Good flag, that was actually on my TODO list. I should have mentioned it's out of scope for this PR, sorry. I was going to handle draft/live once I'd chatted with Dan.

src/server/views/homepage.html
alexluckett: I'll add the wiring in this PR though, I missed it. Thank you.

alexluckett:
> Feels quite AI-written but probably mainly due to the comments.

I will happily say that it is, but I'm still in the driving seat telling it what to do. It's not AI designed though - I can't let it have all the fun. Claude in particular is overly verbose in both its code and comments and has an awful writing style, I usually resolve all of that but I might have missed some comments - will take a look later.

src/client/stylesheets/_service-header.scss
alexluckett: It was a last minute Sonar fix that I let Claude handle, it took the liberty of dumping a novel in.... 🤦

src/server/views/components/service-header/template.njk
alexluckett: It's a lift and shift out of the OneLogin header component repo. The only reason I had to do this was because there were some hardcoded URLs. I don't actually plan on keeping this file longer term, it'll be deleted once we've switched to govuk-frontend v6 and I can use the latest header component.

I didn't change anything with the code simply because it works and visually everything is fine. I think we're okay, I can't see that it would confuse anyone but someone curious looking through the source code.

